PR #7, opened 2025-09-09
Features - Project Member & Account Management

jiramarc:


app/dataset/app/model/project_member.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    field :project_id, type: String, readonly: true
+    
+    field :account_id, type: Integer, readonly: true
+    field :name, type: [String, NilClass]
+    field :email, type: String
+
+    field :role, type: String
+
+    field :invited_by_id, type: Integer
```
anykeyh: readonly true

app/dataset/app/model/project_member.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    field :id, type: Integer, primary: true
+
+    field :project_id, type: String, readonly: true
+    
+    field :account_id, type: Integer, readonly: true
+    field :name, type: [String, NilClass]
+    field :email, type: String
+
+    field :role, type: String
```
anykeyh: beware if this is set like this, anybody with the right to update will update the role too. I would put it as readonly: true.

readonly: true is a bit misleading, but basically it says "do not allow by default to change it when updating". You can still update the field with manual business code.

app/dataset/app/service/project_member/service.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module ProjectMember
+  class Service < Verse::Service::Base
+    use project_members: ProjectMember::Repository
+
+    def index(filter = {}, included: [], page: 1, items_per_page: 1000, sort: nil, query_count: false)
+      project_members.index(
+        filter,
+        included: included,
```
anykeyh: all of this you can use the writing from ruby 3.2:

```ruby
project_members.index(filter, included:, page:, ...)
```
No need to repeat the variable everytime.

app/iam/app/model/account.rb
```diff
@@ -0,0 +1,27 @@
...
+  class Record < Verse::Model::Record::Base
+    type Resource::Iam::Accounts
+
+    field :id, type: Integer, primary: true
+
+    field :name, type: String
+    field :email, type: String
+
+    field :hashed_password, type: [String, NilClass], visible: false, readonly: true
+    field :sso_channel, type: [String, NilClass]
```
anykeyh: readonly

app/iam/app/model/account.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    field :enabled, type: [TrueClass, FalseClass]
```
anykeyh: TrueClass only. There is a weird thing in Ruby as Boolean have two type (TrueClass, FalseClass) but we use only TrueClass as "Boolean"

app/iam/app/model/account.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    field :joined_at, type: [Time, NilClass]
```
anykeyh: readonly

app/iam/app/model/account.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Account
+  class Record < Verse::Model::Record::Base
+    type Resource::Iam::Accounts
+
+    field :id, type: Integer, primary: true
+
+    field :name, type: String
+    field :email, type: String
```
anykeyh: I don't think we want to allow the user to change its email once registered? We should set it to readonly: true too.

anykeyh (changes requested):

anykeyh (approved):

PR #16, opened 2025-09-23
IDAH - Single & Multiple Select DataSource Component Implement

FirstKarnravee:


app/frontend/src/lib/components/app/forms/fields/select/single/single-select-datasource-field.svelte
```diff
@@ -163,13 +172,11 @@
...
                   {@render slotChoice({ choice })}
                 {:else}
-                  <CommandItem onclick={() => select(choice)}>
-                    <CheckIcon
-                      class={cn("mr-2 size-4", {
-                        "opacity-0": choice.value != value,
-                      })}
-                    ></CheckIcon>
-
+                  <CommandItem class="flex w-full gap-0" onclick={() => select(choice)}>
```
jiramarc: you can remove `gap-0` because of by default the `gap` is 0

app/frontend/src/lib/components/app/forms/fields/select/single/single-select-datasource-field.svelte
```diff
@@ -163,13 +172,11 @@
...
+                    {#if choice.value == value}
```
jiramarc: It would be better if you do like this:
```
<CheckIcon class={cn("mr-2 size-4", {
  "opacity-0": choice.value != value
})}
```

because of if we use `{#if}` block the indentation will not align the same across all choice item.

app/frontend/src/lib/components/app/forms/form-field.types.ts
```diff
@@ -26,7 +26,10 @@ export interface SelectFieldBaseProps extends FormFieldBaseProps {
   choices: LabelValue<string | number>[];
   searchable?: boolean;
   searchPlaceholder?: string;
+  searchValue?: string;
   clearable?: boolean;
+  searchKeyWithOperation?: string | null;
```
jiramarc: I think searchKeyWithOperation cannot be `null` as we need to use it for index to key and filter choices

app/frontend/src/lib/components/app/forms/fields/select/single/single-select-datasource-field.svelte
```diff
@@ -120,14 +129,14 @@
       {:else}
         <Button
           variant="outline"
-          class="w-full justify-between"
+          class="w-full justify-between truncate"
```
jiramarc: remove `truncate` class here as you already defined in `<span class="truncate">`

app/frontend/src/lib/components/app/forms/fields/select/single/single-select-datasource-field.svelte
```diff
@@ -22,17 +23,20 @@
...
-    value = $bindable(null),
+    value = null,
     name,
     label,
     placeholder = "Select an option",
     searchable = false,
     searchPlaceholder = "Search an option",
+    searchValue = $bindable(""),
     clearable = false,
+    searchKeyWithOperation = null,
```
jiramarc: This would not be `null`, every devs that use this `<SingleSelectDatasourceField>` should passing a props `searchKeyWithOperation`.

app/frontend/src/lib/components/app/forms/fields/select/single/single-select-datasource-field.svelte
```diff
@@ -90,6 +94,11 @@
     /** Set default sort */
     if (!listOpts.sort) listOpts.sort = ["-id"];
 
+    if (searchable && searchValue) {
+      if (!listOpts.filters) listOpts.filters = {};
+      listOpts.filters[searchKeyWithOperation || "name__match"] = searchValue;
```
jiramarc: Remove `name__match` here as `searchKeyWithOperation` will be defined.

app/frontend/src/lib/components/app/forms/fields/select/single/single-select-datasource-field.svelte
```diff
@@ -3,6 +3,7 @@
 
   import Button from "@/components/ui/button/button.svelte";
   import { Command, CommandEmpty, CommandGroup, CommandItem, CommandList } from "@/components/ui/command";
+  import CommandInput from "@/components/ui/command/command-input.svelte";
```
jiramarc: Move `CommandInput` import into `import { Command, CommandEmpty, CommandGroup, CommandInput, CommandItem, ... } from "@/components/ui/command";`

jiramarc (changes requested):

PR #68, opened 2025-10-15
IDaH - Review System 

Yraimtj:


rcserrano06:
Task linked: [CU-869ar508q IDAH - Review System](https://app.clickup.com/t/869ar508q)

app/dataset/db/migrations/20251015000000_replace_created_by_id_with_email.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+Sequel.migration do
```
anykeyh: Do not create migration for now, add into the existing db schema, as long as the app is not in beta mode.
Also, add a trigram index whenever you use a match operator

anykeyh (changes requested):

app/dataset/app/service/note_feed/service.rb
```diff
@@ -38,13 +38,16 @@ def resolve(id)
...
+      # put created_by_email to nil for now, will be replaced with auth_context[:email] later
+      attr[:created_by_email] ||= nil
+
       attr[:status] = "pending"
 
       entry_id = attr[:entry_id]
       entry = entries.find!(entry_id)
 
-      unless entry.wf_step == "review"
+      unless entry.wf_step == "review" && entry.status == "in_progress"
```
anykeyh: We cannot do that like this.
The idea of the workflow is to be able to parametrize multiple steps.
You need instead to qualify the steps. You cannot just check the name of the step here. What if there is no step called "review"?

app/dataset/db/migrations/20250714000000_create_initial_schema.rb
```diff
@@ -69,10 +69,12 @@
...
     create_table(:projects) do
       column :id, :uuid, primary_key: true, default: Sequel.lit("uuid_generate_v7()")
-      column :name, String, null: false, index: true
+      column :name, String, null: false
 
       column :description, String, null: true
-      column :created_by_id, :bigint, null: true, index: true
+      column :created_by_email, String, null: false, index: true
+
+      index :name, opclass: :gin_trgm_ops, type: :gin
```
anykeyh: Your match operation is on created_by_email but you put your index over `name` column. Doesn't make sense.

anykeyh: I don't understand this index. I can see you did right in other part of the migration, but here you select name. Reasoning? I don't think we want to use match on names. Maybe starts_with but in this case the btree index is okay.

anykeyh (changes requested):

app/dataset/db/migrations/20250714000000_create_initial_schema.rb
Yraimtj: for this is for project name__match app/dataset/app/expo/projects_expo.rb:11 not for created_by_email. But i did not add the created_by_email__match so didn't add this index.

anykeyh (approved):

PR #72, opened 2025-10-20
Login system

Yraimtj:


rcserrano06:
Task linked: [CU-869ar5051 IDAH - Login System](https://app.clickup.com/t/869ar5051)

app/iam/app/expo/auth/helper_methods.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    def set_cookies(auth_token, refresh_token)
+      set_refresh_cookie(refresh_token)
+      set_auth_cookie(auth_token)
+    end
+
+    def set_refresh_cookie(value)
+      if value
+        server.response.set_cookie(
+          Settings["refresh_token.name"],
```
anykeyh: Don't use a setting object. We might want to use a conf in verse config (?)

app/iam/app/expo/auth/simple_expo.rb
```diff
@@ -0,0 +1,86 @@
...
+      input do
+        field(:email, String).filled
+        field(:password, String).filled
+        field?(:cookie, TrueClass).meta(
+          description: <<-MD
+            If true, set the auth and refresh token as cookies
+            in the return response.
+          MD
+        )
+        field(:user_agent, String).meta(description: "The user agent of the client.")
```
anykeyh: This is not something the customer is setting as a field !!!!!!!!!!!!!!!!!!!!!!
This come from the browser, through headers. Remove the field, fetch from ENV variable. Fix nginx if necessary to forward the headers.

app/iam/app/model/account.rb
```diff
@@ -13,15 +13,39 @@ class Record < Verse::Model::Record::Base
...
   class Repository < Verse::Sequel::Repository
     self.table = "accounts"
     self.resource = Resource::Iam::Accounts
+
+    def login(email, password)
+      account = scoped(:login).where(email:).first
+
+      return nil unless account
+
+      # TODO: In Verse => return create_record(account, included: ["active_roles", "person"])
```
anykeyh: you copy paste comments from Pulse which have NO REASON to be in IDAH code. Not good

app/iam/app/model/account.rb
```diff
@@ -13,15 +13,39 @@ class Record < Verse::Model::Record::Base
...
   end
 
   class Repository < Verse::Sequel::Repository
     self.table = "accounts"
     self.resource = Resource::Iam::Accounts
+
+    def login(email, password)
+      account = scoped(:login).where(email:).first
+
+      return nil unless account
```
anykeyh: Small security concern here, as the login will reject quickly if the email is not in the database, while slowly if the password doesn't match. We need to prevent this, probably by matching two invalid passwords or put a sleep.

app/iam/app/model/account_auth.rb
```diff
@@ -0,0 +1,30 @@
...
+    type Resource::Iam::AccountAuths
+
+    field :id, type: Integer
+
+    field :email, type: String
+    field :name, type: String
+    field :picture_url, type: String
+
+    field :role_name, type: String
+    field :role_labels, type: Array
```
anykeyh: useless

app/iam/app/model/account_auth.rb
```diff
@@ -0,0 +1,30 @@
...
+    field :scope, type: Hash
+    # field :auth_type, type: String
+    field :auth_token, type: String
+
+    field :user_agent, type: String
```
anykeyh: useless

app/iam/app/model/account_login_audit.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module AccountLoginAudit
```
anykeyh: AccountAudit(
id: auto, account_id: ref, role (why not), created_at. action (login|logout|change password etc... in the future), ip, user_agent)

anykeyh: - BRIN index on timestamp.
I think also, the account_email should be used. The reason is that we should not have a delete>cascade nor hard reference between user and the audit. In case we delete later a user, we should be able to still have his history.

app/iam/app/model/account_session.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  class Repository < Verse::Sequel::Repository
+    self.table = "account_sessions"
+    self.resource = Resource::Iam::AccountSessions
+
+    def account_login_audit_repo
+      @account_login_audit_repo ||= AccountLoginAudit::Repository.new(auth_context)
+    end
+
+    def check_seq(account_id, session_id, nonce, sequence)
```
anykeyh, line 23: This code is not so robust with multi-refresh token. We need to improve on it a little bit; 
In the case an attacker is using a refresh token, it might receive perpetually an access token in the case he's refreshing fast enough, without the final user noticing it.
I need to think about it.

app/iam/db/migrations/20250714000002_create_account_sessions_and_account_login_audits.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      column :role, String, null: true
+      column :logged_in_at, DateTime, default: Sequel.lit("NOW()"), null: false
+
+      column :ip, String, null: true
+      column :user_agent, String, text: true
+
+      index [:account_id, :logged_in_at]
+      index :logged_in_at, type: :brin # Useful because table is read-only append-only.
+      Migration::Timestamps.timestamps(self)
```
anykeyh: no timestamp; rename logged_in_at to created_at.

app/iam/db/migrations/20250714000002_create_account_sessions_and_account_login_audits.rb
```diff
@@ -0,0 +1,49 @@
...
+      column :user_agent, String
+
+      column :refresh_seq, :bigint, null: false, default: 0
+      column :nonce, :bigint, null: false, default: 0
+
+      Migration::Timestamps.timestamps(self)
+    end
+    Migration::Timestamps.trg_updated_at(self, :account_sessions)
+
+    create_table(:account_login_audits) do
```
anykeyh: `account_action_audit`, cf my comment above.

app/iam/app/model/account_session.rb
Yraimtj: update this but not yet sure as 100%

rcserrano06:
Task linked: [CU-869akbx48 IDAH - Role System](https://app.clickup.com/t/869akbx48)

anykeyh (approved):

PR #73, opened 2025-10-31
Update entry status on processor done

Yraimtj:


anykeyh (changes requested):
- One job = one entry, no need for chunked_index etc...
- You can mark as errored|completed with the same method
- use `processing_error` and `ready` for the status name in the entry. This status might be used for the workflow too, we should discuss further about it.
- I don't think currently the entry is getting the job_id once the job is created, as I inverted control so this would not work in production
- Need to handle case when there is no job created e.g. no processor; the entry should be ready if no job.

anykeyh (approved):

PR #116, opened 2025-11-19
Fix Styling of selected frame on timeline table

FirstKarnravee:
- Fix the styling for the selected frame and the hovered frame.
- Improve the function that checks the current frame while the video is playing.


rcserrano06:
Task linked: [CU-869b5ufxn Annotation timeline zooming](https://app.clickup.com/t/869b5ufxn)

PR #118, opened 2025-11-20
Enable note feed creation at annotate stage & Fix issue when create new annotation note

jiramarc:
…e new annotation note


rcserrano06:
Task linked: [CU-869b7vkhb Enable note feed creation at annotate stage](https://app.clickup.com/t/869b7vkhb)

PR #170, opened 2025-12-18
plugin tools disabled if not wf_step annotate

brian-west-ingedata:


brian-west-ingedata:
I'm really confused on how I keep getting those prettier CI errors while never touching a single css line.

rcserrano06 (approved):

PR #186, opened 2026-01-07
IAM, Dataset & Frontent: improve project members and accounts relation

rcserrano06:
Updates:
- Project member is only disabled when deleted from project.
- Disabling account will automatically disable all project membership.
- Adding project member which account is disabled will automatically enable the account.
- Cancel of pending account invitation will delete the account and project memberships permanently.
- Org owner, Project owner, Reviewer and Annotator can only see accounts with enabled project membership.
- Entries now used project member to view account details.


rcserrano06:
Task linked: [CU-869bnhje5 IDAH - Project member soft deletion](https://app.clickup.com/t/869bnhje5)

PR #191, opened 2026-01-09
Common: fix API token expiration condition

rcserrano06:


rcserrano06:
Task linked: [CU-869bpzhg2 IDAH - Error across IDAH](https://app.clickup.com/t/869bpzhg2)

PR #206, opened 2026-01-20
Submit entry redirects to blank entries list page

jiramarc:
### Issue

Users are redirected to the Entries page when there are no more assigned entries, resulting in a blank page. Occasionally, they may see an empty Entries page when they no longer have permission to access it.


rcserrano06:
Task linked: [CU-869bv0bd1 IDAH - Submit entry redirects to blank Entries page](https://app.clickup.com/t/869bv0bd1)

rcserrano06 (approved):

PR #210, opened 2026-01-22
Add Hover Card on Account Role

jiramarc:
## Summary

- Add HoverCard on AccountRoleBadge to display organizations for each account is an org onwer.
- Add HoverCard on AccountRoleBadge to display projects & project member role for each account is a user.
- Add `disabledActiveFilterSortKeys` props to hide the active filter or sort state for the specific key.
- Update key in `fields` when fetch data with `list()` and `get()` method to use Record.type instead of hard coded string.
- Add `fields` in datasource-table listOptions on each implemented pages.
- Move toast position from top-right to bottom-center


rcserrano06:
Task linked: [CU-869bvhchk IDAH - Display organizations of org owner on Accounts page](https://app.clickup.com/t/869bvhchk)

rcserrano06 (approved):

PR #220, opened 2026-01-30
Annotation Split

sorakrit-audi:
- Split an annotation at the selected frame
  - if split at frame 50, first annotation ends at frame 50 and second annotation starts at frame 51
  - if split frame is the last frame, now it will still split, creating at new frame (may be we should just do nothing here ?)
  - nothing happens if user try to split outside frame range of the annotation
- Interpolate frames for either split annotations if no keyframe is available
- Now has a shortcut with `s` key when an annotation/frame is selected 
- Can be undo with `ctrl-z`


rcserrano06:
Task linked: [CU-869bqpv43 IDAH - Annotation Split](https://app.clickup.com/t/869bqpv43)

app/frontend/src/plugins/idah-video/video-annotation-activity.svelte
```diff
@@ -651,6 +657,181 @@
...
+        context.annotations.delete(newId);
+
+        if ($entryRoot?.metadata.id == newId) $entryRoot = undefined;
+      },
+      isCombinable: () => false,
+      combine: (c) => c,
+    };
+  });
+
+  function getInterpolatedShape(frames: VideoFrameSelection[], frame: number): Point[] | null {
```
brian-west-ingedata: we might want to share definition with the one in svg layer? (once we have polygon/rotation/split consolidated to avoid unnecessary conflict) (to preserve angle/shape specifics algorithm)

sorakrit-audi: @brian-west-ingedata interpolation is now moved/refactored and used by both video-annotation-activity and the svg layer

also now it should also handle rotation interpolation properly with angles

app/frontend/src/lib/command/CommandManager.ts
```diff
@@ -18,15 +18,17 @@ const CommandManager = {
     return this.commands[this.currentIndex];
   },
 
-  add(command: Command) {
+  add(command: Command | undefined) {
```
brian-west-ingedata, line 21: what's the use to add undefined command ?

app/frontend/src/plugins/idah-video/video-annotation-activity/VideoAnnotationContext.ts
```diff
@@ -66,3 +66,37 @@ export interface LabellingConfiguration {
   properties: PropertyConfiguration[];
   taggings: TaggingConfiguration[];
 }
+
+export function getInterpolatedFrame(
```
brian-west-ingedata, line 70: What's the difference between getInterpolatedFrame and currentShape ?

app/frontend/src/lib/command/CommandManager.ts
sorakrit-audi: this is to handle quick returns in commands which has no Command object to return as expected

app/frontend/src/plugins/idah-video/video-annotation-activity/VideoAnnotationContext.ts
sorakrit-audi: will be refactored

PR #251, opened 2026-03-06
Plugin: set decoding/encoding threads for video processing

rcserrano06:


rcserrano06:
Task linked: [CU-869cc6dnb IDAH - Media is not completely process because of server memory full](https://app.clickup.com/t/869cc6dnb)

PR #269, opened 2026-03-16
API keys management system

Yraimtj:


rcserrano06:
Task linked: [CU-869cafkwv IDAH - API key management system](https://app.clickup.com/t/869cafkwv)

common/lib/role_record.rb
```diff
@@ -5,6 +5,10 @@ class RoleRecord < Verse::Model::Record::Base
 
   field :name, type: String, primary: true
 
+  field :mask, type: String do
```
anykeyh: Do we really need mask attribute here? I know it comes from Pulse

anykeyh (changes requested):
Please remove the mask field; it won't be used in this application. You can keep the SemanticVersioning classs in the lib/common

common/lib/role_record.rb
Yraimtj: ok ok.  i'll remove it

PR #278, opened 2026-03-23
Refactoring commands and shortcuts

sorakrit-audi:
Refactored
- `commands` are refactored and now separated from `video-annotation-activity`
- `shortcuts` are refactored, registrations are now generalized and do not have to call different modes separately, just define the mode shortcuts and add to the registry
- `video-annotation-activity` are refactored accordingly to above changes

Additional implementation
- polygon shortcuts


rcserrano06:
Task linked: [CU-869cett90 IDAH - Refactor shortcut/command to be generalized across all plugins](https://app.clickup.com/t/869cett90)
Task linked: [CU-869cfaaun IDAH - Move idah-video from frontend](https://app.clickup.com/t/869cfaaun)

app/frontend/src/lib/shortcut/KeyMapBuilder.ts
```diff
@@ -17,7 +17,7 @@ type ActionKey = string;
 type KeyCombination = string;
 type Action = () => void;
 
-class Builder {
+export class Builder {
```
jiramarc: Should we rename this to `KeyMapBuilder` to be more understandable?

plugins/idah-video/frontend/src/idah/shortcut/key-map-builder.ts
```diff
@@ -17,7 +17,7 @@ type ActionKey = string;
 type KeyCombination = string;
 type Action = () => void;
 
-class Builder {
+export class Builder {
```
jiramarc: Should we rename this to `KeyMapBuilder` to be more understandable?

PR #313, opened 2026-04-02
Plugin: image plugin setup

rcserrano06:


rcserrano06:
Task linked: [CU-869c68e7x Setup for image plugin](https://app.clickup.com/t/869c68e7x)
Task linked: [CU-869cfaaun IDAH - Move idah-video from frontend](https://app.clickup.com/t/869cfaaun)

plugins/idah-image/frontend/src/lib/types.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+export interface IImageAnnotationShape {
+  type: string;
+  start: number;
+  end: number;
+  /** Keyframe selections. */
+  frames: IImageFrameSelection[];
+  /** Allow extensibility. */
+  [key: string]: unknown;
+}
```
brian-west-ingedata: this should not be necessary from the idah-image plugin ?

shape should probably be closer to something like
```
export interface IImageAnnotationShape {
  type: string; # this one should eventually be in the root of annotation even not shape but thats another subject
  points: [number, number][];
  angle: number;
  /** Allow extensibility. */
  [key: string]: unknown;
}
```

brian-west-ingedata: https://github.com/idah-ai/idah/pull/459

PR #320, opened 2026-04-02
Dataset Duplication

rcserrano06:


rcserrano06:
Task linked: [CU-869bjwphd IDAH - Duplicate dataset](https://app.clickup.com/t/869bjwphd)

app/frontend/src/lib/components/app/datasets/overlays/dataset-duplicate-modal.svelte
```diff
@@ -0,0 +1,257 @@
...
+  import { DialogTitle } from "@/components/ui/dialog";
+  import type { EntryRecord } from "@/data/model/dataset/entries/record";
+  import type { FormModalBaseProps } from "@/components/app/overlays/modals/form-modal.types";
+  import { showToast } from "@/components/ui/toast/index.svelte";
+
+  import Button from "@/components/ui/button/button.svelte";
+  import Checkbox from "@/components/ui/checkbox/checkbox.svelte";
+  import EntrySelectionCard from "@/components/app/datasets/entries/cards/entry-selection-card.svelte";
+  import FormModal from "@/components/app/overlays/modals/form-modal.svelte";
+  import InputField from "../../forms/fields/input/input-field.svelte";
```
jiramarc: fix import path properly, so when files is changing the path or directory, it will auto update import path.

app/frontend/src/lib/components/app/datasets/overlays/dataset-duplicate-modal.svelte
```diff
@@ -0,0 +1,257 @@
...
+
+  let isDatasetEmpty = $derived(datasetEntryRecords.length === 0);
+  let isSubmitDisabled = $derived(isDatasetEmpty || selectedEntryIds?.length === 0);
+
+  // set default name when modal opens
+  $effect(() => {
+    if (open) {
+      newDatasetName = `${datasetName} - duplicated`;
+    }
+  });
```
jiramarc, line 62: use `$derived` rune instead of $effect

app/frontend/src/lib/components/app/datasets/overlays/dataset-duplicate-modal.svelte
```diff
@@ -0,0 +1,257 @@
...
+    <div class="grid grid-cols-2 gap-2 overflow-y-auto pr-2 md:grid-cols-4">
+      {#each datasetEntryRecords as entry (entry.id)}
+        <EntrySelectionCard
+          {entry}
+          selected={selectedEntryIds?.includes(entry.id) ?? false}
+          onToggle={handleEntryToggle}
+        />
+      {/each}
+
+      {#if datasetEntryRecords.length === 0}
```
jiramarc: This can be re-write like this
```
{#each datasetEntryRecords as entry (entry.id)}
  ...
{:else}
  <div class="text-muted-foreground col-span-full rounded-md border py-4 text-center text-sm">
     No entries found in this dataset.
  </div>
{/each}
```

jiramarc (changes requested):

app/dataset/app/service/entry/service.rb
```diff
@@ -181,5 +182,58 @@ def unassign_account_entries(account_id, project_id)
...
       end
     end
+
+    def duplicate_entries(dataset_id, duping_dataset_id:, entry_ids: nil, with_annotations: false)
+      duping_entries = if entry_ids
+                         # INFO: also filter with duping_dataset_id to ensure the entries are in the duping dataset
+                         system_entries_repo.index({ id: entry_ids, dataset_id: duping_dataset_id })
+                       else
+                         system_entries_repo.index({ dataset_id: duping_dataset_id })
+                       end
```
rcserrano06, line 210: should use `chunked_index` as `index` is paginated

app/dataset/app/service/entry/service.rb
```diff
@@ -181,5 +182,58 @@ def unassign_account_entries(account_id, project_id)
...
+
+        # Always emit the created event. The media service decides whether the
+        # resource needs processing or is already processed (duplicated media
+        # shares the original's resource) and skips re-running the processor,
+        # so we no longer special-case media readiness here.
+        entries.create(attributes)
+
+        next unless with_annotations
+
+        entry_annotations = annotations.index({ entry_id: duping_entry.id })
```
rcserrano06: should use `chunked_index` as `index` is paginated

app/dataset/app/service/entry/service.rb
```diff
@@ -181,5 +182,58 @@ def unassign_account_entries(account_id, project_id)
...
+          reviewed_by_id: nil,
+          created_at: now,
+          updated_at: now,
+        }
+
+        # Always emit the created event. The media service decides whether the
+        # resource needs processing or is already processed (duplicated media
+        # shares the original's resource) and skips re-running the processor,
+        # so we no longer special-case media readiness here.
+        entries.create(attributes)
```
rcserrano06: I think we should just skip the event on create here? as duplicating an entry means that it's media has completely processed already and has been attached to the entry.

One thing on the frontend is that we should disable the "Duplicate" menu and on the backend prevent duplicate if media is still processing.

app/media/app/service/processor/job.rb
```diff
@@ -7,6 +7,23 @@ class Job < Jobs::Base
...
+      # is skipped. Gated on a *completed* sibling job so retries of a partially
+      # processed resource still run.
+      jobs = Jobs::Repository.new(Verse::Auth::Context.new)
+      if jobs.processed_resource?(resource, except_job_id: job_id)
+        Verse.logger&.info {
+          "[Processor] Resource #{resource} already processed; " \
+            "skipping reprocessing for entry #{@entry_id}"
+        }
+        return
+      end
```
rcserrano06: We can remove this if we skip the event on create of entries.

app/dataset/app/model/dataset.rb
```diff
@@ -243,6 +243,11 @@ def pending!(dataset_id, progress)
       end
     end
 
+    event(name: "duplicated")
+    def duplicated(dataset_id, duping_dataset_id:, entry_ids: nil, with_annotations: false)
+      # NOTE: just publish the event
+    end
```
rcserrano06, line 249: rename method to `duplicate` though event name can be `duplicated`

PR #329, opened 2026-04-07
IDAH - Label configuration property visibility setting

FirstKarnravee:
…just related parsing logic


rcserrano06:
Task linked: [CU-869ctgn71 IDAH - Label configuration property visibility setting](https://app.clickup.com/t/869ctgn71)

brian-west-ingedata (approved):
-const visibility = parser.parse(e.currentTarget.value);
+const visibility = parser.parse(e.currentTarget.value.trim());

PR #342, opened 2026-04-10
IDAH - Workflow registry and workflow selection

Yraimtj:
- add endpoint to list available workflow from default and plugins
- update simple_review_annotation workflow to have workflow definition
- add workflow registration so can get custom workflow plugins
- add workflow_name on dataset table
- update column table so can filter entry by workflow steps that came from workflow config
- can select wished workflow on dataset creation
- add custom submit button on the frontend step


rcserrano06:
Task linked: [CU-869cr63pk IDAH - Workflow registry and workflow selection](https://app.clickup.com/t/869cr63pk)

app/setting/app/expo/plugins_expo.rb
```diff
@@ -51,6 +51,21 @@ def show_modality
     service.show_modality(params[:modality_name])
   end
 
+  expose on_http(
```
anykeyh: I don't understand why the workflows are under plugin?

app/setting/app/service/plugins/service.rb
```diff
@@ -96,6 +96,75 @@ def show_modality(modality_name)
       { shapes: }
     end
 
+    def show_workflows
+      workflows = []
+
+      # Add the default workflow first
+      workflows << {
```
anykeyh: NO

anykeyh: Doesn't make any sense. This is because you've put workflow under plugin. I have no idea why you did that.

anykeyh (changes requested):

app/dataset/app/model/dataset.rb
```diff
@@ -31,7 +32,13 @@ class Record < Verse::Model::Record::Base
     has_many :note_feeds, repository: "NoteFeed::Repository", foreign_key: :dataset_id
 
     def entry_workflow
-      Workflow::SimpleReviewAnnotationWorkflow
+      # Get workflow by workflow_name from the dataset
+      # Falls back to default workflow if no workflow_name or workflow not found
+      if workflow_name.nil? || workflow_name == "default"
+        return Workflow::SimpleReviewAnnotationWorkflow
```
anykeyh: Just add the default workflow into the registry...

app/dataset/app/service/workflow/registry.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module Workflow
+  module Registry
+    extend self
+
+    @workflows = {}
+
+    Entry = Data.define(:plugin, :class_name)
```
anykeyh: Why do you want to store a String as class_name? Why not directly store the workflow class?

app/dataset/app/service/workflow/service.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module Workflow
+  class Service < Verse::Service::Base
+    def list_workflows
+      workflows = []
+
+      # Add workflows from plugins
+      PluginSystem.registry.plugins.each do |plugin_name, plugin|
```
anykeyh: Plugin register workflow on start, not under a list workflows???

app/dataset/app/service/workflow/service.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module Workflow
+  class Service < Verse::Service::Base
+    def list_workflows
```
anykeyh, line 5: Why you don't just list what's inside the registry???

anykeyh (changes requested):

PR #343, opened 2026-04-16
IDAH - Keep original file name for the entry

Yraimtj:


rcserrano06:
Task linked: [CU-869cxybk4 IDAH - Keep original file name for the entry](https://app.clickup.com/t/869cxybk4)

anykeyh (changes requested):
- Use `name` not filename. Because for other modalities (e.g. NLP) you might want to use it while it's not related to a file you a processing.
- Default should not be empty string; could be `nil`. In which case we display the id for example.

anykeyh:
export to UPD should also add the name into the metadata of the entry.

app/frontend/src/lib/components/app/datasets/entries/dropdown-menus/entry-dropdown-menu.svelte
```diff
@@ -109,7 +109,7 @@
       $refetches.entries.list = new Date();
       showToast.success({
         title: "Entry deleted",
-        description: `The entry "${entry.resource}" has been deleted.`,
+        description: `The entry "${entry.name}" has been deleted.`,
```
anykeyh: Above you write:
`entry.name || entry.id`

Then now you assume entry.name is never null. You need to figure out whether it's nullable or not;

anykeyh (changes requested):

anykeyh (approved):

PR #374, opened 2026-04-30
IDAH - Allow to duplicate keyframe

Yraimtj:


rcserrano06:
Task linked: [CU-869d3chd1 IDAH - Allow to duplicate previous keyframe](https://app.clickup.com/t/869d3chd1)

anykeyh (changes requested):
Won't work; we will discuss about it.

PR #379, opened 2026-05-07
Improve/Simplify entry status and workflow step

sorakrit-audi:
**Entry Status Improvement**
- `Pending` now represents the entry waiting for an action, either to be picked up by a background job to process or to be assigned to some member
- `In Progress` now represents the entry being in work, basically currently assigned to some member
- As a result, status `Ready` is deprecated and removed
- Improve and stabilize entry status check loop
- Entry now proceeds onwards for the workflow step once completed with media processing, moving past the `start` stage
- Add some missing methods/endpoints which should be used, ex. `unassign` instead of using `update`

**Related Change**
- Add guard to possible status change of completed entry on select
  - [Completed entry change back to In Progress](https://app.clickup.com/t/869dknw62)


rcserrano06:
Task linked: [CU-869d1h8jx IDAH - Improve entry status display](https://app.clickup.com/t/869d1h8jx)

app/dataset/db/migrations/20260507155600_change_wf_step_default_to_annotate.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+Sequel.migration do
+  change do
+    alter_table(:entries) do
+      set_column_default :wf_step, "annotate"
```
rcserrano06: I think we should not do this on `entries` table as `annotate` is a specific workflow step for idah-video plugin?

rcserrano06 (changes requested):

app/dataset/db/migrations/20260507155600_change_wf_step_default_to_annotate.rb
sorakrit-audi: revised, this is now outdated, no migration changes going forward and this is reverted

PR #382, opened 2026-05-13
IDAH - Polygon feature bugs/improvements

Yraimtj:
- adding capability to scale up/down selected polygon
- when select a group of point in a polygon shape, on click shift and put on top of one point instead of removing only this one point remove all selected point, 
    - if polygon length <= 3 can not delete it, 
     - if interpolated polygon : can not delete all point, but if remove some point => will create shape on frame
- update polygon cursor on creation when hover the first point
- change cursor to crosshair on creation
- add more distinction between existing points and points that can be added.
- fix radius check when create polygon


rcserrano06:
Task linked: [CU-869d8x77g IDAH - Polygon feature bugs/improvements](https://app.clickup.com/t/869d8x77g)

rcserrano06:
Task linked: [CU-869d9ae2z IDAH - Editor Viewport improvements](https://app.clickup.com/t/869d9ae2z)

rcserrano06 (approved):

PR #389, opened 2026-05-15
IDAH - Polygon feature - update cursor

Yraimtj:
Polygon
* change  imaginary point to a solid point to make it more different from other points.
* When creating polygon, its border and points are smaller
* update add and removed cursor
Bounding Box
* Change the point styling to be the same as Polygon
* When the bounding box is selected, its border is same as when not selected 

Both 
- When creating a new annotation (by clicking a tool from top bar), the Properties are not yet validated. Now I can create annotation without entering required properties.
- When selecting the tool (not category) to create annotation, I can still create it without selecting a category.


rcserrano06:
Task linked: [CU-869d8x77g IDAH - Polygon feature bugs/improvements](https://app.clickup.com/t/869d8x77g)

rcserrano06 (changes requested):
Please use also the updated point design on the center of the polygon scale and make the target cursor black when hovering the center circle to be more visible.

Also ensure that we only use 1px line drawing the polygon and bounding box when not selected and selected.

rcserrano06 (approved):

PR #393, opened 2026-05-19
fix: improve category sidebar

FirstKarnravee:
**Left sidebar:**
- [x] Action icons (hide/lock/delete) shouldn’t appear in create/edit mode


rcserrano06:
Task linked: [CU-869d9adq6 IDAH - Editor Category sidebar bugs/improvements](https://app.clickup.com/t/869d9adq6)

rcserrano06 (changes requested):
Should not use the opacity instead add {#if} condition to know if we render the block as it's causing issue with tooltip;

Image

rcserrano06 (approved):

PR #397, opened 2026-05-20
Timeline fixes & improvements (2026-05-19 feedback)

jiramarc:
## Summary

This PR fixes four bugs in the video annotation editor reported by QA.

### Bug fixes

**1. Video speed dropdown opens on play/pause shortcut**
The playback speed dropdown was receiving focus via `bits-ui`'s default focus behavior, causing it to open whenever the play/pause keyboard shortcut was triggered. Fixed by setting `preventScroll` focus option on the speed dropdown trigger to prevent the library's default focus handling.

**2. Deleting the first annotation in a group moves the group to the last row**
The `compareGroups` sort function was ordering groups by `startFrame` first, then `groupId`. When the first annotation in a group was deleted, the group's `startFrame` changed, causing it to jump to a different row position. Fixed by comparing `groupId` first (stable identifier) before `startFrame`, so group row order remains stable regardless of which annotation is deleted.

**3. "Go to frame" input resets to `1` when all digits are deleted**
The `NumberField` input handler in `VideoController` was coercing an empty string to `1` (the field minimum), preventing users from clearing the field before typing a new frame number. Fixed by allowing the field to remain empty until the user commits a value, only clamping on blur/submit.

**4. Timeline track blocks render incorrectly after undo → redo on split annotations**

After the sequence create → extend → split (3–4×) → undo → redo, timeline blocks would overlap or appear as one merged block.

**Root cause:** `annotation.split` was generating a new UUID for the right-side annotation on every `do()` call. After an undo/redo cycle, subsequent split commands still referenced the old right-side ID. When redo ran, `upsert` couldn't find the old ID and inserted a ghost annotation alongside the newly-created one — producing overlapping blocks.

**Fix:** Pre-generate `rightId = uuidv7()` once at command creation time and pass it explicitly to every `create()` call, making the ID stable across all redo cycles.

**Related improvements in the same fix:**
- `{#each}` in `_Track.svelte` now keys by `item.rawData.id` (stable annotation ID) instead of array index, so Svelte correctly reconciles DOM nodes when the list size changes.

## Files changed

| File | Change |
|---|---|
| `VideoController.svelte` | Fix speed dropdown focus; fix "Go to frame" empty-input handling |
| `annotation.ts` | Fix group sort order — compare `groupId` before `startFrame` |
| `annotation/split.ts` | Pre-generate stable `rightId` to fix ghost annotation on redo |
| `_Track.svelte` | Key `{#each}` by annotation ID for correct Svelte reconciliation |

## Test plan

- [x] Play/pause via keyboard shortcut — speed dropdown must not open
- [x] Change playback speed, then play/pause via shortcut — speed must hold, dropdown must not open
- [x] Delete the first annotation in a multi-annotation group — group must stay in the same row
- [x] Type a frame number in "Go to frame", delete all digits one by one — field must become empty, not reset to `1`
- [x] Create annotation → extend → split 3–4× → Ctrl+Z (undo) → Ctrl+Shift+Z (redo) — blocks must render at correct separate positions, no overlap
- [x] Rapid Ctrl+Z / Ctrl+Shift+Z mashing — blocks must remain correct, no ghost annotations

## Other related changes included 

- [x] Update Selection Panel (right sidebar) to use the same sorting as Timeline's, fixing incorrect ordering issue in some cases

### DIFF
--- a/plugins/idah-video/frontend/src/idah/v2/types.ts
+++ b/plugins/idah-video/frontend/src/idah/v2/types.ts
@@ -429,9 +429,6 @@ export interface ICommandDriverV2 {
   /** Listen for palette open/close changes. Returns unsubscribe function. */
   onPaletteChange(cb: (open: boolean) => void): () => void;
 
-  /** Listen for undo/redo stack changes (after call, undo, or redo). Returns unsubscribe function. */
-  onStackChange(cb: () => void): () => void;
-
   // ── Keyboard resolution ────────────────────────────────────────────────
 
   /**
--- a/plugins/idah-video/frontend/src/idah/v2/types.ts
+++ b/plugins/idah-video/frontend/src/idah/v2/types.ts
@@ -429,6 +429,9 @@ export interface ICommandDriverV2 {
   /** Listen for palette open/close changes. Returns unsubscribe function. */
   onPaletteChange(cb: (open: boolean) => void): () => void;
 
+  /** Listen for undo/redo stack changes (after call, undo, or redo). Returns unsubscribe function. */
+  onStackChange(cb: () => void): () => void;
+
   // ── Keyboard resolution ────────────────────────────────────────────────
 
   /**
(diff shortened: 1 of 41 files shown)


rcserrano06:
Task linked: [CU-869d9a7vy IDAH - Timeline bugs/improvements - Continue After Refactor Video Editor](https://app.clickup.com/t/869d9a7vy)

rcserrano06 (approved):

PR #400, opened 2026-05-20
Fix frame step functionality & timeline dark mode

jiramarc:


rcserrano06:
Task linked: [CU-869d8xu4c IDAH - Driver Implementation](https://app.clickup.com/t/869d8xu4c)

plugins/idah-video/frontend/src/lib/commands/viewport/skip-backward.ts
```diff
@@ -9,7 +9,8 @@ import { ui } from "$lib/state/ui.svelte";
 
 function makeAction(): ICommandAction {
   const current = viewport.video.currentFrame.value;
-  const step = ui.frameStep;
+  const localStorageFrameStep = localStorage.getItem("idah-video:settings:frame-step");
```
brian-west-ingedata, line 12: isnt the ui.frameStep already suppose to be coming from localStorage ?

jiramarc: Yes, but we need to get localStorage again when command is trigger. 
Otherwise the `step = UI.frameStep;` will still 10 (default value).

As the annotation top bar is set the value via localstorage but can not set with `ui.frameStep = e.currentTarget.value;` because `ui.frameStep` is the state inside plugin not idah-frontend core.

PR #403, opened 2026-05-21
IDAH - Editor - Fix Annotation Overlay to Reflect Property Styling

Yraimtj:


rcserrano06:
Task linked: [CU-869dcvrev IDAH - Editor - Fix Annotation Overlay to Reflect Property Styling](https://app.clickup.com/t/869dcvrev)

plugins/idah-video/frontend/src/routes/mock/ast-evaluator.ts
```diff
@@ -191,13 +191,9 @@ export class AstProcessor {
       return first;
     }
     // Operator node
-    if (typeof first === "string") {
+    if (typeof first === "string" && this.operators.has(first)) {
```
rcserrano06: Why we change the condition? should this AST Evaluator works the same as the old version as its
not changed?

plugins/idah-video/frontend/src/lib/components/App/Viewport/Shapes/PolygonShape.svelte
```diff
@@ -30,6 +31,14 @@
 
   let color = $derived.by(() => resolveAnnotationColor(annotation));
 
+  // ── Shape display styles from property options ─────────────────────────
+  let shapeStyles = $derived.by(() => resolveShapeStyles(annotation));
+  let shapeStyleString = $derived.by(() => {
```
rcserrano06: it would be good to just return the string value on "resolveShapeStyles" method directly as we only use "shapeStyles" once.

plugins/idah-video/frontend/src/routes/mock/mock-driver.ts
```diff
@@ -394,7 +394,7 @@ export class IdahDriverV2 implements IIdahDriverV2<IVideoAnnotationShape, IVideo
           required: true,
           visibility: [
             "in",
-            [["get", ["value.category"]], [["vehicles/car", "vehicles/bus", "vehicles/van", "vehicles/truck"]]],
+            [["get", ["category"]], ["vehicles/car", "vehicles/bus", "vehicles/van", "vehicles/truck"]],
```
rcserrano06: Why "value.category" is changed? what should be the equivalent AST string of this?

rcserrano06 (changes requested):

plugins/idah-video/frontend/src/routes/mock/ast-evaluator.ts
Yraimtj: tested it with the previous logic and so the wheel property with "get" condition did not show on the sidebar.

plugins/idah-video/frontend/src/routes/mock/mock-driver.ts
Yraimtj: tested it with the previous logic and so the wheel property with "get" condition did not show on the sidebar.

rcserrano06 (approved):

PR #404, opened 2026-05-21
IDAH - Left Sidebar improvement

FirstKarnravee:
Bugfix:
- The actions in the left sidebar affect both shapes when using the same category ID.


rcserrano06:
Task linked: [CU-869dchqgd IDAH - Editor - Hide/Lock/Delete Function Affects Both Bounding Box and Polygon Annotations with Same Category ID](https://app.clickup.com/t/869dchqgd)

plugins/idah-video/frontend/src/lib/components/App/CategorySelector/menus.ts
```diff
@@ -14,28 +14,38 @@ export interface CategoryAction {
...
+
   getDriver().command.call("annotation.toggle_category_editability", {
     category: categoryId,
+    shapeType,
   });
 }
 
-export function deleteCategoryAnnotations(categoryId: string) {
+export function deleteCategoryAnnotations(opts: { categoryId: string }) {
+  const { categoryId } = opts;
```
rcserrano06: We should consider also `shapeType` as the same category can exists but different shape type, like:
```
bounding box
  - vehicles
     - car

polygon
  - vehicles
     - car
```

rcserrano06 (changes requested):

plugins/idah-video/frontend/src/lib/commands/category/delete.ts
```diff
@@ -68,8 +69,8 @@ export function register(driver: IIdahDriverV2): void {
         categoryAnnotations = props.annotations;
       } else if (props.category) {
         // Resolve annotations from category tree
-        categoryAnnotations = data.annotations.items.filter((ann) =>
-          isCategoryMatch(ann.value?.category, props.category),
+        categoryAnnotations = data.annotations.items.filter(
+          (ann) => isCategoryMatch(ann.value?.category, props.category) && ann.shape.type === props.shapeType,
```
rcserrano06: I think you need to separate the condition for the `props.shapeType` filter as if it's `null` then it will not return any annotations, what I recommend:
```typescript
} else if (props.category || props.shapeType) {
  let filterAnnotations = data.annotations.items;

  if (props.category) {
    categoryAnnotations = filterAnnotations.filter((ann) =>
      isCategoryMatch(ann.value?.category, props.category),
    );
  }

  if (props.shapeType) {
    categoryAnnotations = filterAnnotations.filter((ann) =>
      ann.shape.type === props.shapeType,
    );
  }
}
```
With this we can delete annotations based on `shapeType` or `category`

Implement the same condition with:
```
toggle_editability.ts
toggle_visibility.ts
```

rcserrano06 (changes requested):

rcserrano06 (approved):

PR #421, opened 2026-05-29
IDAH - Restrict upload modal closure to X button only

FirstKarnravee:


rcserrano06:
Task linked: [CU-869dg3c5y IDAH - Restrict upload modal closure to X button only](https://app.clickup.com/t/869dg3c5y)

app/frontend/src/lib/components/app/datasets/overlays/dataset-form-modal.svelte
```diff
@@ -168,6 +168,6 @@
   }
 </script>
 
-<FormModal {action} {title} loading={submitting} onCancel={resetForm} onConfirm={submit} bind:open>
+<FormModal {action} {title} loading={submitting} onCancel={resetForm} onConfirm={submit} canClickOutside bind:open>
```
rcserrano06: Why do we add `canClickOutside` props here if it's not present in the current component? it will seem to be useless as by default it would be false? same as with the components below.

rcserrano06 (changes requested):

app/frontend/src/lib/components/app/overlays/modals/form-modal.svelte
```diff
@@ -28,6 +28,7 @@
     actions,
     confirm,
     children,
+    canClickOutside = false,
```
rcserrano06: This can be set by default to `true` so we expect the form to be closable when clicking outside then just disable it if we want to.

rcserrano06 (changes requested):

rcserrano06 (approved):

PR #425, opened 2026-05-30
Scroll timeline to current frame instead of annotation start

jiramarc:
- Skip scrolling when the current frame is already visible in the viewport
- Center the current frame horizontally when a scroll is needed, preserving zoom scale
- Remove the shape-dependency that blocked the command when no shape was attached


rcserrano06:
Task linked: [CU-869dfgd37 IDAH - Improve timeline navigation when selecting annotation](https://app.clickup.com/t/869dfgd37)

rcserrano06 (approved):

PR #430, opened 2026-06-02
IDAH - Video zoom out is not working on small viewport

Yraimtj:


rcserrano06:
Task linked: [CU-869dh9um1 IDAH - Video zoom out is not working on small viewport](https://app.clickup.com/t/869dh9um1)

plugins/idah-video/frontend/src/lib/components/App/Viewport/Viewport.svelte
```diff
@@ -96,12 +96,16 @@
     viewport.workspace.clampTranslate();
   }
 
+  /** Minimum zoom: allows zooming out to fit even large media in a small viewport. */
+  const MIN_ZOOM = 0.05;
+  const MAX_ZOOM = 100;
```
rcserrano06: It would be good to add as:
```ts
export const MIN_ZOOM = 0.05;
export const MAX_ZOOM = 100;
```
in `plugins/idah-video/frontend/src/lib/state/viewport.svelte.ts` so we can just use it on any component or commands.

rcserrano06 (changes requested):

rcserrano06 (approved):

PR #436, opened 2026-06-08
Plugin: idah-video refactor video buffering and frame navigation

rcserrano06:
### Problem

Video frame navigation was slow, especially on mobile networks. The system was using HLS `currentLevel` to switch quality, which flushed the entire media buffer on every seek. This forced the browser to re-download fragments it already had cached.

### Solution

**1. Use `loadLevel` instead of `currentLevel`**
- Only affects future downloads; preserves buffered fragments instead of discarding them.

**2. Predict-then-verify for slow networks**
- Before attempting high-quality on slow connections, estimate if it can download in ~1 second
- If not, fetch low-quality immediately
- Remember if HQ is too slow ("latch") and skip it on future seeks until the network improves

**3. Stricter step-through gate**
- Prevent arrow-key presses from queuing up and replaying later, which was causing cascading re-seeks
- Safety valves: timeline clicks always work, and a 2-second timeout releases the gate if the network stalls

**4. 15-second upgrade watchdog**
- Cancel quality upgrades that stall, preventing hangs

### Impact

- Instant frame seeking on buffered content; no re-downloads
- Slow networks get low-quality fallback within 1 second
- Better UX on mobile networks


rcserrano06:
Task linked: [CU-869dfzauy IDAH - Fix video buffering issue](https://app.clickup.com/t/869dfzauy)

PR #442, opened 2026-06-10
IDAH - Error when Org Owner upload a video

Yraimtj:
__Root cause:__ `Jobs::Repository#scoped` in `app/media/app/model/jobs.rb` was using `scope_with_media_resources` which fetched all media resources into Ruby first (`media_repo.index({}).map(&:resource).uniq`), then filtered jobs with `IN ?`. When the media repository's own scoping returned an empty set (e.g. a user not found in project memberships, or an empty `IN ()` clause), `find!` would see no rows and raise `RecordNotFound` → 404 — even though the job physically existed in the DB.

__The fix:__ Replaced the `scoped` method in `app/media/app/model/jobs.rb` with the same pattern used in `app/sync/app/model/jobs.rb`:

- __`as_org_owner?`__ — gets project IDs for the user's org via the Dataset API, then uses an `EXISTS` subquery against the `medias` table to check `medias.resource = arguments->>'resource' AND medias.project_id IN :project_ids`
- __`as_user?`__ — gets project IDs from the user's project memberships via the Dataset API, same `EXISTS` subquery

This eliminates the two-step Ruby fetch that was causing the empty-resources problem, and instead handles authorization entirely in SQL — consistent with how the sync service scopes its jobs.


rcserrano06:
Task linked: [CU-869dkf6z6 IDAH - Error when Org Owner upload a video](https://app.clickup.com/t/869dkf6z6)

app/media/app/model/jobs.rb
```diff
@@ -33,9 +33,45 @@ def scoped(action)
...
+          frag = <<-SQL
+            EXISTS (
+              SELECT 1
+              FROM medias
+              WHERE medias.resource = (arguments->>'resource')
+                AND medias.project_id IN :project_ids
+            )
+          SQL
+
+          table.where(Sequel.lit(frag, project_ids:))
```
rcserrano06: Put into private method named:
```
def scope_with_org_projects_medias
```

app/media/app/model/jobs.rb
```diff
@@ -33,9 +33,45 @@ def scoped(action)
...
```
rcserrano06: Put into private method named:
```
def scope_with_member_projects_medias
```

rcserrano06 (changes requested):

rcserrano06 (changes requested):
Add `enabled: true` as memberships can be disabled when fetching:
```
Api[:idah].dataset.project_members.index_all({ ... })
```

Also please check other API requests trying to get project membership and ensure that they have `enabled: true` paramter

app/media/app/model/medias.rb
```diff
@@ -102,14 +102,14 @@ def delete!(id)
 
     def project_from_orgs_scoped
       org_ids = auth_context.custom_scopes[:org]
-      projects = Api[:idah].dataset.projects.index(organization_id: org_ids).data
+      projects = Api[:idah].dataset.projects.index_all(organization_id: org_ids)
```
anykeyh, line 105: DO NOT LOAD IN MEMORY.

anykeyh:
We should not have RPC call in role scoping

anykeyh (approved):

PR #443, opened 2026-06-11
IDAH - API keys visible and editable across users

FirstKarnravee:
- Add a filter to display only the user's keys.

### DIFF
--- a/app/frontend/src/routes/(protected)/(app)/api-keys/+page.svelte
+++ b/app/frontend/src/routes/(protected)/(app)/api-keys/+page.svelte
@@ -135,6 +135,9 @@
             "status",
           ],
         },
+        filters: {
+          account_id: $authStatus.authContext?.id,
+        },
       }}
       {onLoadSetContexts}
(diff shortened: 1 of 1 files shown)


rcserrano06:
Task linked: [CU-869dn9kmy IDAH - API keys visible and editable across users](https://app.clickup.com/t/869dn9kmy)

jiramarc (approved):

app/frontend/src/routes/(protected)/(app)/api-keys/+page.svelte
```diff
@@ -135,6 +135,9 @@
             "status",
           ],
         },
+        filters: {
+          account_id: $authStatus.authContext?.id,
+        },
```
rcserrano06, line 140: I think this should be manage through the backend scopes and not in the frontend, as I see here that user can access API keys which they did not create.

rcserrano06 (changes requested):

PR #452, opened 2026-06-17
IDAH - Project description not displayed on Project page

FirstKarnravee:
## Summary

Fix an issue where the project description was not displayed on the Project page.

## Root Cause

The Project page only requested the project name from the backend and did not include the `description` field in the project fetch request. As a result, the project description was unavailable for rendering in the page header.

## Changes

* Updated the project fetch request to include the `description` field.
* Added project description rendering below the project name in the page header.
* Ensured the description is displayed consistently across Project pages after data is loaded.

## Testing

* Verified that projects with descriptions display the description correctly on the Project page.
* Verified that projects without descriptions continue to load without errors.
* Confirmed that existing Project page functionality remains unchanged.

### DIFF
--- a/app/frontend/src/routes/(protected)/(app)/projects/[projectId]/+layout.svelte
+++ b/app/frontend/src/routes/(protected)/(app)/projects/[projectId]/+layout.svelte
@@ -39,9 +39,7 @@
   // Records
   let project: ProjectRecord = $state(new ProjectRecord());
 
-  $effect(() => {
-    setContext("project", project);
-  });
+  setContext("project", project);
 
   // Lifecycle
   onMount(async () => {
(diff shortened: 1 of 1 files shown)


rcserrano06:
Task linked: [CU-869dqudfc IDAH - Project description not displayed on Project page](https://app.clickup.com/t/869dqudfc)

app/frontend/src/routes/(protected)/(app)/projects/[projectId]/+layout.svelte
```diff
@@ -92,6 +91,7 @@
               <Text size="h2" weight="semibold">{project.name}</Text>
               <ProjectDropdownMenu {projectId} align="center" />
             </div>
+            <Text size="sm">{project.description}</Text>
```
Comment (jiramarc) [inline app/frontend/src/routes/(protected)/(app)/projects/[projectId]/+layout.svelte]: the `project.description` can be null if user doesn't set it; 
So you need to wrap this display project description with `{#if}` block to ensure that we don't render a blank line after title.

jiramarc (changes requested):

jiramarc (approved):

PR #453, opened 2026-06-18
Fix multiple bugs in entries list and entry card

FirstKarnravee:
- Fix sort "none" branch clearing all sorts instead of only the target column
- Fix delete toast using entry.resource instead of entry.name
- Fix unAssignEntries running sequentially; switch to Promise.all
- Fix select-all checkbox missing indeterminate state for partial selection
- Fix export modal using "entries(s)" typo; use pluralizeUnit consistently
- Fix deleteEntries error handler using unsafe error.errors access


rcserrano06:
Task linked: [CU-869dgppaz IDAH - Fix entry order shuffle and search by name filter](https://app.clickup.com/t/869dgppaz)

PR #465, opened 2026-06-19
Fix some commands not working in editor mode

sorakrit-audi:


app/frontend/src/lib/plugin/v2/driver/manager/command-manager.ts
```diff
@@ -23,7 +23,7 @@ export class CommandManagerV2 {
   private combineWindow = 5000;
 
   /** Current driver mode, used by getActiveCommands(). Updated externally. */
-  currentMode: string = "default";
+  currentMode: string = "editor";
```
rcserrano06, line 26: it would be good to use the const variable on:
```
app/media/plugins/idah-image/frontend/src/lib/types.ts:13
```
same with the "review" mode

rcserrano06 (changes requested):

rcserrano06 (approved):

PR #479, opened 2026-06-22
Fix entry-member relation and fix assignment info for non-member admin/org owner

sorakrit-audi:
- entry now correctly look up for member only in corresponding project, previously able to look up across project as it was only using `account_id`
- add `assigned_to_email`, `submitted_to_email`, and `reviewed_to_email` for simplicity in lookup and historical cases where a user is no longer a member, updated along side `assigned_to_id`, `submitted_to_id`, and `reviewed_to_id` respectively
- `admin / org owner` now get assigned and shown correctly without needing to be a member on entry selection


rcserrano06:
Task linked: [CU-869dn7dzq IDAH - Unassigned entry auto-assignment not reflected on frontend for Admin/Org Owner](https://app.clickup.com/t/869dn7dzq)

anykeyh (approved):

PR #491, opened 2026-06-24
Dataset status revision

sorakrit-audi:
- Dataset status should be changed to `In Progress` if there's an assigned entry
- Dataset should stay as `Pending` as an entry is added but not assigned
- Dataset should stay as `In Progress` if entry is submitted and goes into pending to wait for reviewer assignment because at least one of the entries in that dataset has been processed
- Dataset should be changed from `Completed` to `In Progress` if it was completed then a new entry is added (not assigned yet) because at least one of the entries in that dataset has been processed

### DIFF
--- a/app/dataset/app/model/dataset.rb
+++ b/app/dataset/app/model/dataset.rb
@@ -167,13 +167,24 @@ module Dataset
       return if total_entries.zero?
 
       completed_entries = dataset[:entries_completed_count]
+      in_progress_entries = dataset[:entries_in_progress_count]
+
+      # Entries awaiting next-stage assignment (e.g. submitted, waiting for a
+      # reviewer) are back to "pending" status but have already been worked on.
+      # submitted_by_id distinguishes them from fresh, untouched entries.
+      submitted_entries = table.db[:entries]
+                               .where(dataset_id:)
+                               .exclude(submitted_by_id: nil)
+                               .count
 
       progress = completed_entries.to_f / total_entries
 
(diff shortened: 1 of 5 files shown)


rcserrano06:
Task linked: [CU-869dv7jg2 IDAH - Dataset status incorrectly set to In Progres](https://app.clickup.com/t/869dv7jg2)

app/dataset/app/model/dataset.rb
```diff
@@ -167,13 +167,24 @@ def update_progress!(dataset_id)
...
       completed_entries = dataset[:entries_completed_count]
+      in_progress_entries = dataset[:entries_in_progress_count]
+
+      # Entries awaiting next-stage assignment (e.g. submitted, waiting for a
+      # reviewer) are back to "pending" status but have already been worked on.
+      # submitted_by_id distinguishes them from fresh, untouched entries.
+      submitted_entries = table.db[:entries]
+                               .where(dataset_id:)
+                               .exclude(submitted_by_id: nil)
+                               .count
```
rcserrano06: It might be good to update the triggers and add `entries_submitted_count` rather than calling the DB every time.

You can find the trigger at:
```
app/dataset/db/migrations/20250714000000_create_initial_schema.rb:167
```

We will need to update on the server later to correct the submitted count of existing datasets.

rcserrano06 (changes requested):

rcserrano06 (approved):

PR #499, opened 2026-06-29
mutli step shape fixes (history (undo/redo), color,..)

brian-west-ingedata:
Multi-step shape undo/redo, pending shape visuals, and timeline frame-seeking fixes**

- Rewrote `add_point` commands with **snapshot-based do/undo** (captures `snapshotBefore` at creation time, sets mode + frame in the right order) so undo/redo work reliably across mode switches
- Unified `reset()`/`replace()` into a single `set points()` setter on polygon/line drafts
- Added frame-seeking to video `annotation.add`, `keyframe_add`, `keyframe_delete`, `delete`, and `split` commands so undo/redo navigate to the relevant frame
- Used `uuidv7()` pre-generated IDs in video `annotation.add` so redo reuses the same annotation ID for downstream keyframe operations
- Added `hexToRgba()` and `color` prop to all create-shape overlays so category colors show during drawing
- Rendered pending shapes as finalized `AnnotationGeometry` during popover, with category color reacting to `pendingValue`
- Added `onStackChange` to `CommandManagerV2` so the toolbar updates undo/redo state immediately


rcserrano06:
Task linked: [CU-869dw71uz IDAH - Undo not working for polygon point adding in image plugin](https://app.clickup.com/t/869dw71uz)

plugins/idah-image/frontend/src/lib/components/App/Viewport/Shapes/ShapesContainer.svelte
```diff
@@ -464,57 +477,62 @@
 
     <g>
       <!-- Build mode: bounding box creation preview -->
-      {#if isBoundingBoxMode}
+      {#if isBoundingBoxMode && !pendingAnnotation}
```
rcserrano06: Everything is good to me just a question, should we add also this guard `!pendingAnnotation` on the idah-video?

rcserrano06 (changes requested):

plugins/idah-image/frontend/src/lib/commands/annotation/add.ts
```diff
@@ -62,6 +64,16 @@ export function register(driver: IIdahDriverV2): void {
           if (data.annotations) {
             await data.annotations.delete(createdId);
           }
+          // Restore drawing mode for multi-step shapes so the user can
+          // continue editing or individually undo add_point commands.
+          // For line, restore the first point so the preview is visible.
+          if (props.shape.type === IMAGE_POLYGON) {
```
jiramarc, line 70: Would this be better if we use switch case?

plugins/idah-image/frontend/src/lib/components/App/Viewport/Shapes/ShapesContainer.svelte
brian-west-ingedata: I currently went the other way around and removed it. as the drafts shape are now cleared anyway when we have pending Annotation

plugins/idah-image/frontend/src/lib/commands/annotation/add.ts
brian-west-ingedata: equivalent

rcserrano06 (approved):

plugins/idah-image/frontend/src/lib/components/App/Viewport/Shapes/ShapesContainer.svelte
brian-west-ingedata: https://github.com/idah-ai/idah/pull/504 I actually put it back to prevent stale history if clicked outside of the properties pop up

PR #501, opened 2026-06-29
Frontend: fix entry auto select text width

rcserrano06:


rcserrano06:
Task linked: [CU-869dupxjz IDAH - Annotation actions not properly restricted in Review mode](https://app.clickup.com/t/869dupxjz)

PR #505, opened 2026-06-30
Allow reordering of Shapes and Categories in Label Editor

jiramarc:
## Summary

Adds native HTML5 drag-and-drop reordering in the dataset **Label Editor** so teams can organize labels to match their annotation workflow.

- **Categories** — drag a label (by its grip handle) to reorder within its level. Moving a parent node moves its whole subtree; drops are constrained to same-parent siblings. Persisted via the existing `categoryOrderMap` + `sortValuesByTreeOrder` on save (no schema change).
- **Shapes** — drag to reorder the configuration list. Added an optional `order` field to `IShapeConfig` because `labeling_configuration` is a `jsonb` column and Postgres does **not** preserve object-key order — so shape order must live in a persisted value. New/duplicated shapes append at the end.

## Bug fixes included
- Save button stayed disabled after a reorder (change wasn't reflected in `labelConfig`).
- Reordering an unselectable/intermediate node (not in `values`) didn't persist — now moves its descendant block.
- Wrong shape selected by default on page load (used jsonb key order instead of the displayed `order`).

## Files
- `categories/category-tree-node.svelte` — draggable grip, drop handlers, drop-line indicator
- `categories/category-tree.svelte` — drag state + controller
- `label-config-editor.svelte` — shape ordering/selection, shape drag controller
- `plugin/v2/types.ts` — `order?: number` on `IShapeConfig`
- `labels/+page.svelte` — `reorderCategory` (block move), `reorderShape`, wiring

## Verification
- `pnpm check` clean for all touched files (pre-existing repo-wide errors unrelated).
- Manually tested: reorder shapes + categories (incl. nested/intermediate nodes), Save enables, persists after save + reload, correct default selection.

Refs: CU-869dvq3u7


rcserrano06:
Task linked: [CU-869dvq3u7 IDAH - Allow reordering of Shapes and Categories in Label Editor](https://app.clickup.com/t/869dvq3u7)

rcserrano06 (changes requested):
Just one thing I think the shapes should also be sorted on the Editor view when we re-order it from the label editor.

rcserrano06 (approved):

PR #508, opened 2026-07-02
Label Config Template (Frontend)

jiramarc:
## What
Adds the frontend and supporting backend for Label Configuration Templates: users can save a dataset's label configuration as a reusable template, browse/apply templates filtered by modality, rename/delete them, and replace a dataset's config with an existing template — from the dataset Label Editor.

## Why
Today, reusing a label configuration means duplicating it from a specific dataset or rebuilding it from scratch. This lets users save a config once as an org-scoped template and reuse it across any dataset of a matching modality, cutting repetitive setup (CU-869dwrycg).

## How
- Backend: new `LabelingConfigurationTemplate` model, service, and expo (CRUD + migration), registered in `app/dataset/config/routes.rb`
- Extracted `LabelConfigController` (`.svelte.ts`) so label-config state is shared between the dataset Label Editor and the new template management modal, instead of being duplicated
- Added `LabelConfigTemplateDropdownMenu` and `LabelConfigTemplateManagementSheet` for create / inline-rename / delete / save-as-new / replace template flows, with a reusable `EditableTextField` and delete confirmation
- Added modality filtering so only templates matching the dataset's modality are selectable, and consolidated refetch triggers (`refetch.ts`) to keep the sheet and editor in sync after mutations (replacing an earlier derived `sheetRefreshKey`)
- Added `dataset:labeling_configuration_templates` resource and granted it to `admin`, `org_owner`, and `user` roles so template management respects existing role scoping
- Refactored dataset modality badge rendering into a shared `badgeMap` util (`DatasetModalityBadge`), reused by the new template UI

## Testing
- `bundle exec rspec app/dataset/spec` — covers the new model, service, and expo specs
- Manual: in the dataset Label Editor, save the current config as a new template, apply an existing template (verify modality filtering), rename and delete a template, and replace a dataset's config with a template — confirm role restrictions (non-admin/owner cannot manage templates)

---
Closes CU-869dwt0fr


rcserrano06:
Task linked: [CU-869dwt0fr Frontend](https://app.clickup.com/t/869dwt0fr)

rcserrano06:
Task linked: [CU-869dwrycg IDAH - Save Label Configuration as Reusable Template](https://app.clickup.com/t/869dwrycg)

rcserrano06 (changes requested):
Improvements to do:
- Remove file .claude/worktrees/audit-findings
- On overwriting existing dataset and there are changes on the label config that has not been saved yet, it would be best to show error that the user has unsaved changed.

---

- Add confirmation when overwriting an existing template, say that the template will get completely overwritten.
Image

----
 
- When duplicating the current label config on other dataset remove the current dataset on the list.
Image

rcserrano06 (approved):

PR #512, opened 2026-07-03
Plugins: idah-video prefer video stream duration over container duration

rcserrano06:


rcserrano06:
Task linked: [CU-869dzawuj IDAH-VIDEO - issue with duration when importing media](https://app.clickup.com/t/869dzawuj)

PR #516, opened 2026-07-06
Notification and Frontend: scoped audit cleanup

FirstKarnravee:
---
# Audit Remediation — Tiered Summary

Part of the Notification + Frontend audit remediation work.

Findings are grouped by implementation complexity and coordination cost rather than audit severity.

> This PR also includes **one fix outside the audit's 25 findings** — see "Additional fix (outside audit scope)".

---

# ✅ Tier 1 — Delivered in this PR

Low-risk, self-contained changes that can be implemented and reviewed independently.

## Frontend

* [x] **FE-017** — Remove reset token from the URL while preserving the reset flow.
* [x] **FE-018** — Fix audit-log actor avatars and account resource resolution.
* [x] **FE-013** — Remove dead debug logging from the AST resolver.
* [x] **FE-019** — Fix `passowrd` → `password` typo in `AuthService`.
* [x] **FE-009** — Document trusted `{@html}` usage.
* [x] **FE-020** — Document intentional `Object.assign()` mutation.

## Notification Service

* [x] **N-06** — Move sender address to `MAIL_FROM` configuration.
* [x] **N-08** — Replace notification channel literal with shared constant.
* [x] **N-13** — Fix invitation email copy typo.

---

# ⏳ Tier 2 — Follow-up Implementation

Worth implementing, but requires additional environment setup, dependencies, or validation. **All now delivered in this PR.**

* [x] **N-02** — Auto-escape HTML email templates to prevent stored XSS.
* [x] **N-04** — Add SMTP timeouts to prevent worker hangs.
* [x] **N-11** — Add delivery error handling and structured logging.
* [x] **N-01** — Validate `to_email` before sending emails.

---

# 🤝 Tier 3 — Team Decisions Required

Requires architectural or product decisions before implementation. None are blocking Tier 1/Tier 2 delivery.

### [ ] N-14 — Decide how email delivery should be audited

**Current state:**
- The notification service persists nothing of its own — no notification-log table; sent emails leave no durable record.
- N-11 now logs delivery failures, but there is no queryable audit trail of what was sent, to whom, when, and with what outcome.

**Decision required:**
- Pick an approach: (A) structured logging only — cheapest, not queryable; (B) a minimal `notification_logs` table owned by the notification service — self-contained, no cross-service change; (C) emit a synthetic event the `app/audit` service subscribes to — cross-service, currently avoided.

**Reason:**
- `app/audit` is resource-event driven, so it is *not* a drop-in target (that's why option C is cross-service).
- The finding's wording ("notification log *table*") and this codebase's audit convention (DB-backed, e.g. `app/audit`'s `logs` table) lean toward option B.
- A full implementation plan with review-ready code for both A and B exists in `.claude/plan/n-14-notification-audit-trail.md`.

### [ ] Delivery guarantees — Decide retry and dead-letter strategy

**Related findings:**
- **N-11** introduced delivery error handling and logging, but the notification channel still provides no delivery guarantee.
- **N-14** highlights the lack of a durable audit trail, which would also be required for any future retry / dead-letter implementation.

**Current state:**
- The event channel is **at-most-once**: verse-redis reads the stream with `noack: true` (`stream/subscriber/stream.rb:252`) — no pending list, no retry, no reclaim, no dead-letter queue.
- On a handler exception (N-11 re-raises), the message is logged and **dropped**; the re-raise does *not* trigger redelivery. There is no framework-level guarantee to tune.
- Recovery today is user/admin-initiated only: manual "resend invitation" (invites) and re-requesting a password reset. Other notifications are informational.

**Decision required:**
- Whether to *build* any automatic guarantee (the framework provides none). If yes, scope it — e.g. bounded retry + a durable dead-letter / `notification_logs` record for the critical path (password reset).

**Reason:**
- This is **not** a poison-message-loop risk (`noack: true` means nothing loops); the real gap is **silent loss** of transient failures.
- Recommend making failures visible first (N-14), then deciding retry from real failure-rate data rather than speculation.

### [ ] FE-017 trade-off — Confirm the refresh behavior is acceptable

**Current state:**
- FE-017 strips `?token=` from the reset-password URL after capturing it into `$state`.
- Consequence: a browser **refresh** on `/reset-password` loses the token, so a subsequent submit would fail — the user must re-open the reset/invite link.

**Decision required:**
- Confirm this refresh trade-off is acceptable for the single-use invite/reset flow.

**Reason:**
- The strip *is* the security fix (token no longer lingers in history or the `Referer` header). Keeping the token in the URL to survive refresh would re-introduce exactly the leak we're fixing, so the trade-off is inherent. A single-use flow makes a mid-flow refresh uncommon.

---

# 🔧 Additional fix (outside audit scope)

* [x] **IAM — invitation email sent after commit.** The account-creation invitation email is now published in `accounts.after_commit` instead of inside the create transaction. Previously the notification worker's cross-service account lookup ran before the row committed, raised `NotFound`, and dropped the invite (it only arrived via a manual "resend invitation"). Found while testing N-04; **not one of the 25 audit findings** — flagged here so the iam-service change isn't invisible to reviewers.

---

# 📌 Deferred / Out of Scope

These findings are intentionally not implemented and are not considered pending work.

### FE-023 — Cache key ignores auth identity

**Deferred by team decision.** The practical risk is low and a proper fix would require broader cache architecture changes with meaningful regression risk.

### FE-007 — Open redirect via `redirectTo`

**Framework mitigated.** SvelteKit's `goto` rejects cross-origin navigation, so the reported exploit is not reproducible on the current stack.
**Verified:** `/login?redirectTo=//evil.com` and `?redirectTo=https://evil.com` both land on `/projects`, confirming the mitigation.

### FE-014 — Dead `delayed()` helper

**Obsolete.** The reported code was already removed in a previous commit.

### FE-022 — `list({ all: true })` recursion bug

**Deferred.** The bug is real but currently has zero callers and will be addressed before first use.

### N-03 — Account enumeration

**Rejected.** The service consumes trusted internal Redis events and exposes no attacker-facing oracle.

### N-09 — Fail-open notification default

**Intentional product behavior.** Notifications default to enabled unless the user opts out.

### N-10 / N-15 — Template path traversal and generic error handling

**Deferred.** The affected values are internal static strings and are not user controlled.

### N-05 / N-12 — Response caching and SMTP pooling

**Deferred.** Low-value performance optimizations that add complexity with little practical benefit.

### N-07 — Development SMTP without TLS

**Rejected.** Applies only to local development and does not affect production.

---

# Validation Performed

* [x] `pnpm build` — passes.
* [x] `eslint` — clean on the changed frontend files.
* [x] `rubocop` — clean on the notification changes (N-01, N-02, N-04, N-11).
* [x] `svelte-check` — **no new type errors introduced** (106 pre-existing, unchanged vs. baseline; the tool is not green independent of this PR).
* [x] Notification `rspec` — renderer (N-02) + service (N-01/N-11) + expo specs pass **locally** (18 examples, 0 failures). CI to confirm.
* [x] SMTP timeout — standalone smoke against a black-holed host raised `Net::ReadTimeout` at the configured timeout (N-04); the rescue + re-raise path (N-11) is covered by the service spec (not driven end-to-end).

Recommended before merge (not yet run in-browser):

* [ ] FE-017 reset-password browser smoke — token gone from the URL, submit still resets.
* [x] FE-018 audit-logs UI check — avatars render; an `accounts` resource log shows an email, not a UUID.

---

# Completeness Map

* **Tier 1 · Delivered:** 9 findings
* **Tier 2 · Follow-up (all delivered):** 4 findings
* **Tier 3 · Decisions:** 1 finding (+ delivery-guarantees & FE-017 trade-off notes)
* **Deferred / Out of Scope:** 11 findings

**All 25 audit findings are accounted for (implemented, deferred, rejected, or pending team decision).** Plus **1 additional non-finding fix** (IAM invitation email).


rcserrano06:
Task linked: [CU-869e0fx6t Notification and Frontend (well-scoped)](https://app.clickup.com/t/869e0fx6t)

PR #519, opened 2026-07-06
Setting + Frontend core/criticals

jiramarc:
## Audit remediation — Setting + Frontend core/criticals

Remediation for the 38 findings in Marc's audit scope (Setting `S-01…S-25` + 13 core/critical `FE-*`). Findings are tracked below in a **new priority order** — sorted so *easy + important* fixes land first, with architecture / cross-team-blocked work grouped at the bottom.

**This PR implements `FE-008` only** (Critical XSS). The remaining items are tracked here and land in follow-up PRs on this branch.

### FE-008 — what changed
`marked` no longer sanitises its output (since v4), so rendering it through `{@html}` allowed stored XSS via any user-authored markdown. The parsed HTML is now run through `DOMPurify` before rendering. The app is SSR (`adapter-node`) and `dompurify` needs a DOM, so sanitisation happens in the browser only and returns `""` on the server until hydration re-runs the parser.

---

### Tier 1 — Quick wins · easy + important + solo-actionable
- [x] **FE-008** · Crit · XSS: sanitise `marked` output before `{@html}` ← **this PR**
- [x] **S-01** · High · `serve_asset` path-traversal containment
- [x] **S-02** · High · `serve_file` traversal guard *(also resolves S-14, S-19)*
- [x] **S-04** · High* · delete dead `:create` no-op line *(severity overstated — already filtered)*
- [x] **FE-010** · Med · fix `filtersToHash` + add `uri.test.ts` — ⚠️ **audit impact overstated**: the buggy `filtersToHash` is in `utils/uri.ts` and is **unused**; `BackendDataSource.list` uses the *correct* `filtersToHash` from `@/data/filtering`, so live list filtering was never broken. Fixed the dead duplicate + added tests.
- [x] **FE-021** · Med · `roles` prop = **AND** with `can()`, not silent override — audited all 8 `PageProvider` callers: 7 unaffected. `audit-logs` would have regressed (no role had any `audit:logs` right, so `can()` was false for admins), so also granted admins `audit:logs.read.*` in `common/data/roles/8.0.0_admin.yml`. ⚠️ **role-model change: admins need a token refresh / re-login to pick it up.** `<Can>` has the same override bug — fixed under FE-005.
- [ ] **S-13** · Med · clamp `items_per_page` — ⏭️ **Skipped (deferred to Verse):** an unbounded/`1000` default is not specific to `AccountSettings::Service` (the same appears in e.g. audit `Log::Service#index`). A per-service clamp leaves the pattern everywhere else, so the fix belongs in Verse's shared pagination layer, not here. Track as a framework-level follow-up.
- [x] **S-08** · Med · add `plugins (name, version)` unique index (migration) — ⚠️ fails loudly if duplicate rows already exist (dedupe manually before running)
- [ ] **S-12** · Med · remove phantom `::SCHEDULER`/`::EXECUTOR` worker-boot block — ⏭️ **Skipped (infra/config owner):** `puma.rb` worker config should be aligned across services by the deployment/infra owner, not changed ad hoc in one service. Defer to that owner. (Latent unless `PUMA_WORKERS > 1`.)
- [x] **FE-002** · Low · regex → `startsWith` (regex-injection)
- [x] **FE-001** · Low · store + clear the `setInterval` handle
- [x] **FE-012** · Low · `throw parseSingleElementError(...)` instead of a bare string
- [x] **S-16** · Low · `Sequel.cast(...)` instead of raw `Sequel.lit` (both `set` methods; backfill migration's constant-value `lit` left as-is, out of scope)
- [x] **S-18** · Low · typo `curent_plugin_name` → `current_plugin_name`
- [x] **S-20** · Low · use `Resource::Iam::Accounts` constant, not `"iam:accounts"` — spec literal kept intentionally (independent wire-value assertion)

### Tier 2 — Dead-scaffolding cleanup · easy, solo · decision = delete
- [x] **S-23** · Info · delete commented-out `plugin_images` schema block

### Tier 3 — Medium effort, solo · important but not one-liners
- [x] **FE-015** · Crit · validate `pluginId` (allowlist regex `/^[A-Za-z0-9_-]+$/` + `{#if}` head guard). Strategic isolation (iframe/SRI/off-origin) still tracked with S-06/S-07.
- [x] **FE-003** · Med · shared `requestJson` + `rejectionFor` helpers (`response.ok` + defensive JSON parse) across all 5 methods; updated test mocks (also fixed a pre-existing failing `#delete` test) + added a non-ok/non-JSON case
- [x] **FE-005** · Med · reactive `$effect` `<Can>` (fixes stale-on-mount) + AND `roles` (finishes the FE-021 fix for `<Can>`). Verified no effect loop + `authStatus` doesn't emit on navigation.
- [x] **FE-004** · Med · promise-cache `(projectId, accountId)` member lookup on AuthContext (dedupes concurrent `can()`; instance-scoped auto-invalidation; failures not cached)
- [x] **FE-011** · Med · parallelise remaining assign/priority loops (`Promise.all`) + hoist the assign modal's redundant per-iteration email fetch. (Server batch endpoint remains a Tier-5 dataset-team follow-up.)

### Tier 4 — Hygiene · easy but low/Info importance
- [x] **S-22** · Info · rewrite stale FR/EN route comments
- [x] **S-24** · Info · remove `puts last_response.body` from specs
- [x] **S-25** · Info · drop unused `rmagick` + ImageMagick/ffmpeg from image

- [ ] **S-09** · Med · delete broken `Plugins::Manager` — ⏭️ **moved from Tier 2:** deleting a half-built plugin-install subsystem is a delete-or-finish decision (part of the phantom plugin-lifecycle story with S-10/S-11/S-12). Needs senior/team sign-off before removal.
- [ ] **S-21** · Low · undefined `install_plugin` in `Service#install` — ⏭️ **moved from Tier 2:** coupled to S-09 (removed by the same delete). Same senior/team discussion.
- [ ] **S-10** · Med · remove dead `SettingsExpo` route registration — ⏭️ **moved from Tier 2:** `/settings` is intended (but unbuilt) system-settings scaffolding; whether to remove vs design the feature (system-vs-org-vs-user scope + admin auth) needs senior/team sign-off.
- [ ] **S-11** · Med · `PluginLifecycleContext` no-op stub — ⏭️ **moved from Tier 2:** documenting the no-op as intentional vs implementing mount/unmount decides whether the Setting service hosts plugin backends at all — an architecture call needing senior/team sign-off.
- [ ] **S-06** · High · harden plugin asset serving (ext allowlist + size cap + `nosniff` + MIME map) — ⏭️ **moved from Tier 3:** needs the Setting exposition-spec integration checkpoint + touches renderer wiring. Senior discussion pending. Implementation blueprint in the appendix below.
- [ ] **S-17** · Low · replace anonymous inline renderer with tested `PluginMimeType` helper — ⏭️ **moved from Tier 3:** paired with S-06 (same change). Same checkpoint + discussion.
- [ ] **S-15** · Med · extract shared `PluginCatalog` (kill per-request `Dir.glob` in `Service#find` + `Plugin::Record#path`) — ⏭️ **moved from Tier 3:** backend refactor on the request hot path; needs the Setting integration checkpoint to verify. Also the natural home for the S-07 allowlist gate.
- [ ] **S-07** · Med(RCE) · plugin-loader trust — loader in `common/` (Roger). *Solo interim: allowlist gate.*
- [ ] **S-03** · High · anonymous endpoints — needs FE-015 owner sign-off. *Solo interim: gate `modalities`/`show_modality`.*
- [ ] **S-05** · High · event-bus trust boundary — infra/Roger. *Solo interim: log + narrow `delete`.*
- [ ] **FE-016** · Med · remove `Record`'s `[key: string]: any` — **scoped follow-up PR** (measured: surfaces **44 new type errors across ~20 files**, not ~15; several need design decisions). Milestones:
  1. Add contained typed accessor `Record.get<T>(key)` / `setField(key, value)`; drop the index signature; fix the internal `this[recordKey]` in `getValueByKey`.
  2. Migrate mechanical dynamic-access sites (`record[key]` → `record.get<T>(key)`) — combobox/select field components, datasource-table filters, etc.
  3. **Design cluster:** `Property 'resource_id'/… does not exist on type 'T'` (~10) — declare the missing fields on the record classes (proper fix) rather than papering over with `.get()`.
  4. Fix `T[keyof T]` generic typing in the shared select/combobox field components (~8).
- [ ] **FE-006** · Info · auth module singletons — document only, do-not-pursue.
- [ ] **FE-011 (batch)** · Med · server-side bulk endpoint — dataset backend team.

---

### Related findings discovered while fixing (not in original audit)
- **Duplicate `filtersToHash`** — one in `utils/uri.ts` (was buggy, unused) and one in `data/filtering.ts` (correct, used). Candidate for de-duplication so the dead copy can't drift back into use.
- **Dead array branch in `convertSearchParamsToHash`** (`utils/uri.ts`) — it routes through `Object.fromEntries(searchParams.entries())`, which collapses duplicate query keys to the last value before the loop, so its `[value]`-promotion branch is unreachable for `URLSearchParams` input. Behaviour is captured (last-wins) in `uri.test.ts`; logic left unchanged to avoid altering `datasource-table` behaviour out of scope.

Source: `marc-audit-findings.md`. Tier 5 items require coordination (Roger for `common/` loader + event bus, the FE-015 owner, and the dataset backend team).

---

## Appendix — S-06 + S-17 implementation blueprint (ready to execute at the checkpoint)

Both findings are one coherent change on the plugin file/asset endpoints. Deferred to Tier 5 because it needs the Setting exposition-spec integration checkpoint and touches renderer wiring — pending senior discussion.

**1. New util `app/setting/app/util/plugin_mime_type.rb`** (named + testable — resolves S-17's intent): a module mapping extension → safe Content-Type, `DEFAULT = "application/octet-stream"`, method `PluginMimeType.for(filename)` using `File.extname(...).downcase`. Cover `.js .css .json .png .jpg .jpeg .gif .svg .webp .woff .woff2`. Add `plugin_mime_type_spec.rb` (known/unknown/uppercase/no-ext).

**2. Service `serve_asset` (`app/setting/app/service/plugins/service.rb`)** — after the existing S-01 containment + `File.file?` checks, add:
```ruby
ALLOWED_ASSET_EXT = %w[.png .jpg .jpeg .gif .svg .webp .js .css .json .woff .woff2].freeze
MAX_ASSET_BYTES   = 10 * 1024 * 1024
...
return nil unless ALLOWED_ASSET_EXT.include?(File.extname(asset_path).downcase)
return nil if File.size(asset_path) > MAX_ASSET_BYTES
```

**3. Expo (`app/setting/app/expo/plugins_expo.rb`)** — use the `renderer.content_type = …` / `server.response.headers[…] = …` idiom (as in `medias_expo.rb:57`, `exports_expo.rb:35`):
- **`serve` (files):** delete the anonymous `Class.new` renderer → `renderer: Verse::Http::Renderer::Identity`; in the action set `Content-Type = PluginMimeType.for(params[:filename])` + `X-Content-Type-Options: nosniff` (404 if nil).
- **`serve_asset`:** keep `Verse::Http::Renderer::Binary`; set `renderer.content_type = PluginMimeType.for(params[:splat].last)` + `nosniff` before returning the IO.

**4. Extend `plugins_expo_spec.rb`** — assert `Content-Type` + `nosniff` on serve/serve_asset (existing 200/body + 404 assertions must still pass).

**Residual risk:** SVG stays allowlisted (plugins use `icon.svg`); `nosniff` + `image/svg+xml` is set, but an SVG opened via direct top-level navigation can still execute script — belongs to the strategic off-origin/sandbox follow-up (with FE-015/S-07), not this tactical pass.

**Verify:** `bundle exec rspec app/setting/app/util/plugin_mime_type_spec.rb` and `.../plugins_expo_spec.rb` (mock the service, no DB); `ruby -c` the edited files; optionally `curl -i` a `/files/plugin.js` + `/assets/x.png` on a booted service and confirm headers, and that a `.txt`/oversized asset 404s.

Refs: CU-869e0fx5t

### DIFF
--- a/app/frontend/package.json
+++ b/app/frontend/package.json
@@ -63,6 +63,7 @@
   },
   "dependencies": {
     "date-fns": "^4.1.0",
+    "dompurify": "^3.4.11",
     "marked": "^16.4.0",
     "mode-watcher": "^1.1.0",
     "uuidv7": "^1.0.2",
--- a/app/frontend/pnpm-lock.yaml
+++ b/app/frontend/pnpm-lock.yaml
@@ -11,6 +11,9 @@ importers:
       date-fns:
         specifier: ^4.1.0
         version: 4.1.0
+      dompurify:
+        specifier: ^3.4.11
+        version: 3.4.11
       marked:
         specifier: ^16.4.0
         version: 16.4.0
@@ -797,6 +800,9 @@ packages:
   '@types/resolve@1.20.2':
     resolution: {integrity: sha512-60BCwRFOZCQhDncwQdxxeOEEkbc5dIMccYLwbxsS4TUNeVECQ/pBJ0j09mrHOl/JJvpRPGwO9SvE4nR2Nb/a4Q==}
 
+  '@types/trusted-types@2.0.7':
+    resolution: {integrity: sha512-ScaPdn1dQczgbl0QFTeTOmVHFULt394XJgOQNoyVhZ6r2vLnMLJfBPd53SB52T/3G36VI1/g2MZaX0cwDuXsfw==}
+
   '@typescript-eslint/eslint-plugin@8.42.0':
     resolution: {integrity: sha512-Aq2dPqsQkxHOLfb2OPv43RnIvfj05nw8v/6n3B2NABIPpHnjQnaLo9QGMTvml+tv4korl/Cjfrb/BYhoL8UUTQ==}
     engines: {node: ^18.18.0 || ^20.9.0 || >=21.1.0}
@@ -1090,6 +1096,9 @@ packages:
 
   dom-walk@0.1.2:
     resolution: {integrity: sha512-6QvTW9mrGeIegrFXdtQi9pk7O/nSK6lSdXW2eqUspN5LWD7UTji2Fqw5V2YLjBpHEoU9Xl/eUWNpDeZvoyOv2w==}
+
+  dompurify@3.4.11:
+    resolution: {integrity: sha512-zhlUV12GsaRzMsf9q5M254YhA4+VuF0fG+QFqu6aYpoGlKtz+w8//jBcGVYBgQkR5GHjUomejY84AV+/uPbWdw==}
 
   enhanced-resolve@5.18.3:
     resolution: {integrity: sha512-d4lC8xfavMeBjzGr2vECC3fsGXziXZQyJxD868h2M/mBI3PwAuODxAkLkq5HYuvrPYcUtiLzsTo8U3PgX3Ocww==}
@@ -2719,6 +2728,9 @@ snapshots:
 
   '@types/resolve@1.20.2': {}
 
+  '@types/trusted-types@2.0.7':
+    optional: true
+
   '@typescript-eslint/eslint-plugin@8.42.0(@typescript-eslint/parser@8.42.0(eslint@9.35.0(jiti@2.5.1))(typescript@5.9.2))(eslint@9.35.0(jiti@2.5.1))(typescript@5.9.2)':
     dependencies:
       '@eslint-community/regexpp': 4.12.1
@@ -3034,6 +3046,10 @@ snapshots:
   dom-accessibility-api@0.6.3: {}
 
   dom-walk@0.1.2: {}
+
+  dompurify@3.4.11:
+    optionalDependencies:
+      '@types/trusted-types': 2.0.7
 
   enhanced-resolve@5.18.3:
     dependencies:
--- a/app/frontend/src/lib/components/app/datasets/entries/overlays/_AssignEntryFormModal.svelte
+++ b/app/frontend/src/lib/components/app/datasets/entries/overlays/_AssignEntryFormModal.svelte
@@ -34,22 +34,20 @@
   async function assignMember(): Promise<void> {
     if (entryIds.length === 0 || !selectedMemberAccountId) return;
 
-    let projectMemberRecord: ProjectMemberRecord | undefined = $state(undefined);
-    for (const entryId of entryIds) {
-      await entriesBackendDataSource.assign({ id: entryId, memberAccountId: selectedMemberAccountId });
+    const memberAccountId = selectedMemberAccountId;
 
-      /** Get the email of selected member */
-      const projectMemberRes = await projectMembersBackendDataSource.list({
-        fields: {
-          [ProjectMemberRecord.type]: ["email"],
-        },
-        filters: {
-          account_id: selectedMemberAccountId,
-        },
-      });
+    await Promise.all(entryIds.map((entryId) => entriesBackendDataSource.assign({ id: entryId, memberAccountId })));
 
-      projectMemberRecord = projectMemberRes.data.at(0);
-    }
+    /** Get the email of the selected member (identical for every entry). */
+    const projectMemberRes = await projectMembersBackendDataSource.list({
+      fields: {
+        [ProjectMemberRecord.type]: ["email"],
+      },
+      filters: {
+        account_id: memberAccountId,
+      },
+    });
+    const projectMemberRecord = projectMemberRes.data.at(0);
 
     open = false;
     selectedMemberAccountId = null;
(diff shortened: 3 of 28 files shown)


rcserrano06:
Task linked: [CU-869e0fx5t Setting and Frontend (core/criticals)](https://app.clickup.com/t/869e0fx5t)

PR #527, opened 2026-07-06
Sync and Audit UPDATE

Yraimtj:


rcserrano06:
Task linked: [CU-869e0fx4w Sync and Audit](https://app.clickup.com/t/869e0fx4w)

common/lib/api/exposition.rb
```diff
@@ -216,6 +230,61 @@ def calculate_multipart_length(params, boundary)
...
+      use_ssl = uri.scheme == "https"
+      http.use_ssl = use_ssl
+
+      # TODO: It's probably better to generate a set of self-signed authority
+      # internally to the cluster., rather than disabling verification.
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE if use_ssl
+
+      timeout = options[:timeout] || 300
+
+      Timeout.timeout(timeout, "Request timed out after #{timeout}s") do
```
anykeyh: Do not use timeout, that's not good; use the option timeout in the http request itself;

The Timeout method is a trap, as it can lead to memory leak (file descriptor never collected)

anykeyh (changes requested):

PR #530, opened 2026-07-07
Frontend & Plugins: idah-image enhance annotation fetching

rcserrano06:
## Batch / progressive annotation loading

Speeds up loading large annotation sets and renders them progressively
instead of blocking until the entire dataset is ready.

### What changed

**Streaming fetch**
- `fetch()` now accepts an optional `onBatch` callback. Records are delivered
  in chunks as they load — each backend page on a cold cache, or as the local
  IndexedDB cursor walks on a warm cache — so the UI can paint incrementally.
- Increased sync page size from **100 → 1000** annotations per request.
- Cold-start streaming returns early and skips the redundant full IDB re-read,
  since every record was already delivered via `onBatch`.

**Idempotent bulk append**
- Added `bulkAppend()` to the data store, backed by a `Set<string>` membership
  index for O(1) dedup — replaces the old O(n²) merge scan.
- Idempotent by design, so overlapping/retried batches (e.g. a warm delta
  re-emitted by the full read) are skipped with no extra bookkeeping.
- The annotation store now paints each streamed page as it arrives.

**UI**
- New `LoadingIndicator` component showing load progress, wired into the Viewport.
- Updated `ShapesContainer`, `SyncIndicator`, and `ImageAnnotationWorkspace`
  for progressive rendering and improved `visibleAnnotations` sorting.


rcserrano06:
Task linked: [CU-869e14ap3 IDAH - Improve annotation loading on editor initial load](https://app.clickup.com/t/869e14ap3)

PR #531, opened 2026-07-08
Dataset and Media revision

sorakrit-audi:
# Dataset + Media Code Revision

Scope: the dataset (`D-xx`) and media (`M-xx`) findings from the assigned audit
docs (`audit/03-dataset-service.md`, `audit/04-media-service.md`,
`audit/00-ASSIGNMENTS.md`). Recheck fold-ins (`R-xx`) and issues found during
implementation (`FI-xx`) are listed separately at the end — they are not part of
the assigned audit set.

---

## What have been done

Issues addressed and committed, from the dataset/media audit docs.

### Obvious Findings

| Item | Description | Type | Approach |
|------|-------------|------|-------------|
| D03 | `note_feed/service.rb` copied the annotation without checking its `entry_id` matched the feed's — a client could anchor an annotation from a different entry (cross-entry IDOR). Spec covers a mismatched `entry_id` → rejected | fix + spec | Reject a note-feed annotation whose `entry_id` doesn't match the feed's entry |
| D04 | `entry.rb` / `entry/service.rb` used `find_by!`, so a duplicate or late Redis job event on an already-transitioned entry raised. Switched to `find_by` + early return. Spec covers a repeated event | fix + spec | Make `mark_entries_status_as` idempotent so duplicate/late job events are no-ops |
| D05 | `project_member/service.rb#update` had no field whitelist (any attribute writable) and its owner-role guard only checked promotion. Added `UPDATABLE_FIELDS` `.slice` + a bidirectional `project_owner` guard. Spec both directions | fix + spec | Whitelist member-update fields; fire the owner-role guard on either direction |
| D09 | New migration `20260709000000_...`; `annotations.updated_at`, `note_feeds[project_id, status]`, `entries.resource` were unindexed. Plain indexes (not unique — dataset duplication deliberately reuses `resource`) | migration | Add indexes on hot filter columns |
| D10 | `entry_stats/recompute.rb#persist` built the repo with `Repository.new(nil)`; replaced with `Verse::Auth::Context[:system]`. Regression spec constructs the repo the way persist does | fix + spec | Use the system auth context in EntryStats persist instead of `nil` |
| D13 | `entry/service.rb#assign_member` assigned to any account with a membership row, including soft-disabled ones. Added a `disabled_at`-nil check. Spec: disabled member → rejected | fix + spec | Require an active (non-disabled) project membership to assign an entry |
| D22 | `status`/`wf_step`/`job_id` were freely PATCHable via the API (`entry.rb` / `entry/service.rb`). Guard strips them unless the caller resolves to `:all`. Not `readonly` — media legitimately PATCHes them with a system token, which `readonly` would silently drop | fix + spec | Service-layer guard stripping `status`/`wf_step`/`job_id` from non-system writes |
| D26 | `annotations_expo_spec.rb` helper posted to a wrong route and was unused; the real `"batch"` path is already covered | deletion | Drop the dead `rpc_batch_call` spec helper |
| M-10 | `jobs/scheduler.rb#process` did `raise e` after `jobs.error(...)`, re-raising into a pool that already rescues+logs — pointless. Dropped it; added explicit backtrace logging where the error is handled | fix + spec | Stop re-raising into the already-logging thread pool |
| M-11 | Bypassed the logger and leaked resource ids to stdout; service is dead until `Video::Job` (M-01) exists, so deleted rather than converted | deletion | Remove debug `puts` in `Video::Service` |
| M-12 | `processor/context.rb#download_original` did `tempfile.write(file.read)`, loading the whole remote object into a String. Switched to `IO.copy_stream(file, tempfile)` | fix + spec | Stream media download instead of slurping into memory |
| M-14 | `jobs/scheduler.rb` called `jobs.reschedule(job.id, scheduled_at:)` with a stray positional → `ArgumentError`. Fixed the signature. Had to land before M-17 | fix + spec | Fix the `reschedule` call signature in the scheduler |
| M-17 | `processor/context.rb` delegated to `job.reschedule!`, which doesn't exist; fixed to `job.reschedule(after:)`. Spec uses `instance_double` so the nonexistent method is verified against the real class | fix + spec | Fix `Context#reschedule!` delegating to a missing bang method |

### Items worked on with considerations

| Item | Description | Type | Approach |
|------|-------------|------|-------------|
| D02 | `annotation/service.rb:54-58` overwrites `project_id`/`dataset_id`/`entry_id` from the scoped entry — the *only* barrier, since the JSON:API create schema doesn't strip readonly fields. No exploit today; risk is a refactor dropping the block. Spec + load-bearing comment. **Evidence:** mutation removing the block persisted the client-supplied id | spec + comment | Regression spec pinning the annotation-create id reassignment |
| D06 | `entry.rb#select` ran its claim UPDATE under the broad read scope after a separate `status == "pending"` check, so two racers both passed and the second overwrote the first. Narrowed the UPDATE scope to `status pending AND (assigned_to_id nil OR self)`; a 0-row update returns false → raise. **Evidence:** red steal-test raised nothing pre-fix, green after | fix + spec | Atomic entry claim; loser gets a clear "already claimed" error |
| D16 | `entry.rb#assigned` and `project_member.rb#enabled` coerce with `value.to_s.downcase == "true"`, so `"1"`/`"yes"` fall to the false branch. Behavior kept (FE sends `"true"`/`"false"`); specs assert exact result sets so any coercion change is deliberate | specs | Pin the `assigned`/`enabled` filter string-coercion contract |
| D18 | `simple_review_annotation_workflow.rb:30-35` `sample_rate \|\| 1`. The skip path (`annotate → done`) was already tested at rate 0 (`service_spec.rb:562`), but the `\|\| 1` default (key absent) was not. Added a default-fallback spec + doc comment. **Evidence:** mutation `\|\| 0` landed the entry in "done" | spec + doc | Cover the untested `sample_rate` default; document 0..1 semantics |
| D23 | `dataset/service.rb:39` commented-out assignment; `datasets` has no such column (initial schema has it on projects/annotations/note_feeds/note_comments only) and `Dataset::Record` declares no such field — so it couldn't be uncommented without a migration | deletion | Remove dead `created_by_email` comment on dataset create |
| M-09 | `medias/service.rb:9` injected `Video::Service` as `video_service`, never referenced (grep: only other mention is its own spec), falsely implying uploads trigger video processing. Class untouched | deletion | Remove dead `use_system video_service` injection |
| M-16 | `jobs/base.rb#emit` routes through `@command.call`, but `@command` is set only inside `run(&block)`; emitting earlier gave `NoMethodError: undefined method 'call' for nil`. Guarded with a self-explanatory message. New `base_spec.rb`. **Evidence:** mutation removing the guard reproduced the exact NoMethodError | guard + spec | Clear error when `emit` is called before `run` |
| M-20 | `set_keys.rb` accepts raw PEM or `file:/path`; a bare path is parsed as key content → opaque OpenSSL error. The initializer is **byte-identical across all 7 services** (same md5), so documented in `config/keys/README.md` rather than diverge one copy | doc | Document `SERVICE_JWT_KEY` `file:` convention |
| M-22 | `processor/service.rb:34-55` loops a modality's processors and overwrites `entry.job_id` each iteration, so with >1 only the last links. **Evidence:** only `idah-image` and `idah-video` register, one processor each → last-wins unreachable today. Replaced the TODO with the assumption + data-model pointer | comment | Document the single-processor-per-modality `job_id` assumption |
| M-23 | `medias.rb:62-63` warned that actor-less background-processor media events cause audit error logs. The audit guard `logs_expo.rb:156` (`return unless actor_account_id`) already skips them. **Evidence:** git timeline — the guard (2026-01-22) postdates the TODO (2026-01-15) by a week | deletion | Remove the stale actor-less-media-event TODO |

---

## What to be considered

Real items left open — deferred deliberately, worth doing later.

| Item | Description | Type | Why deferred / what's needed |
|------|-------------|------|------------------------------|
| D19 | `annotations_expo.rb:38-49` (`RpcCreateSchema`, `batch_limit: 50`): `dimensions`/`annotation`/`metadata` are open `jsonb` Hashes with no size/depth limit, ×50 per batch. In-app there is no cap today | hardening | Not needed now; needs a per-field byte cap + confirmation of the ingress body limit |
| M-13 | `medias.rb:103-115` and `jobs.rb:126-164` scoped queries call `dataset.projects/project_members.index_all` — 1–2 synchronous HTTP hops to Dataset per query. Amplifies latency and couples Media availability to Dataset (Dataset down ⇒ all Media reads error) | architecture | Intended for now; measure reads/sec + p95 before choosing cache vs. denormalization |
| M-15 | `jobs.rb:116-122` — `select(Sequel.lit("min(scheduled_at)")).first[:min]` relies on Postgres naming the aggregate column `min`; works but brittle to a Sequel/PG change | cleanup | Correct today; do the explicit alias only when next editing `jobs.rb` (e.g. for M-13) |
| M-21 | `jobs_expo.rb:3` extends `Verse::Exposition::Base` directly, not `BaseExpo`; works via its own `json_api` block but won't inherit future `BaseExpo` cross-cutting config | consistency | Align when next touching the file, verifying response parity |

---

## What have been skipped

Decided not to act on — settled design decision, not a real problem, falsified,
owned elsewhere, or accepted trade-off.

### Won't-fix (verified intended / falsified / owned elsewhere)

| Item | Description | Why not |
|------|-------------|---------|
| D01 | Audit flagged that an annotator can review their own submission | Intended for solo use; per-workflow `allow_self_review` flag only if team separation is ever needed |
| D07 | Audit flagged a note-feed `:resolve` scope mismatch | Code already matches intended resolve policy; flag came from a stale doc-comment (optional 1-line doc fix) |
| D08 / M-18 | Audit claimed `page[size]` was unbounded (large-page DoS vector) | **Falsified** — framework already caps `page[size]` at 1000 (`verse-jsonapi dsl/index.rb:80-83`) |
| D11 | `entry_stats/recompute.rb:15-27` — the `rescue StandardError` around plugin stats also swallows `StatKeyConflictError` (a plugin *registration* bug), with only a log line for visibility | Log-only is sufficient; no publish/event wanted (re-raising would also drop core stats) |
| D12 | Audit flagged EntryStats' system-context use as a naming/scoping smell | Legitimate system-context use; the real underlying bug is R-01 (tracked below) |
| D14 | Audit flagged a project_member issue already resolved in code | Already fixed at `project_member.rb:67` — verify post-merge only |
| D15 | Audit flagged a per-submit read as a potential N+1/perf concern | No bulk-submit path; single indexed read per submit is negligible |
| D17 | Audit flagged unused multi-workflow scaffolding | Dead multi-workflow scaffolding; roadmap call |
| D20 | `dataset/service.rb:61-82` — `notify_dataset_completed` loops project owners and publishes one notification each on the handler thread. Publish is already async (Redis); cost is N publishes + a member query | Negligible at single-digit owner counts; a real fix needs a new batched cross-service contract |
| D21 | `project_member/service.rb:47-63` — `after_commit` calls `accounts.show` (sync IAM) to gate the email; if IAM is down the member commits but the notification is lost, no retry | Straightforward and retriggerable; the proposed job-based fix doesn't fit (dataset has no job system) |
| D24 | Audit flagged org-owner note moderation as scoped to `own` rather than org-wide | **Policy call, not a bug** — `permission_spec.rb` doc-tables document org-owner note update/delete as `own`/`own` and code matches; needs explicit product decision |
| D25 | Audit flagged `updated_at` propagation causing extra writes | Accepted write-amplification trade-off |
| M-07 | Audit flagged `use_system` on a background-job service | Intended for system-triggered background jobs |
| M-08 | Audit flagged the unused `EXECUTOR` injection | Owned by Roger's Executor consolidation — verify his change lands |
| M-19 | `config/initializers/scheduler.rb:13` — unconditional `puts ENV["PUMA_WORKERS"]` on boot, bypassing the logger | Left as-is by decision |

---

## Appendix — Recheck fold-ins and issues found during implementation

Not part of the assigned dataset/media audit set. Surfaced during our own
verification pass (`R-xx`) or while implementing the
commits above (`FI-xx`). Deferred to a follow-up.

### Recheck fold-ins (self-discovered, not in the assigned 43)

| Item | Description | Type | Summary |
|------|-------------|------|---------|
| R-01 | `scoped_query/service.rb:10-18` gates create-authorization for annotations, note feeds, entries, datasets, and members by role membership but not `disabled_at IS NULL` — a soft-removed member keeps creating records. One-line SQL fix, matches every other scope | authz gap | `ScopedQuery::Service.with_project_access?` ignores `disabled_at` |
| R-02 | `note_feed.rb:158-207` — the `:resolve` branch's two `project_members` subqueries lack the `disabled_at` check that `:read` and `:update/:delete` have; a soft-removed owner/reviewer can still resolve. Same root cause as R-01 | authz gap | Note-feed `:resolve` scope omits `disabled_at` |
| R-03 | `jobs/scheduler.rb:79-86` — `constantize(job.job_class)` + subclass check run in the run-loop *before* the thread pool's rescue; a bad `job_class` raises straight into the loop, killing the scheduler thread. The job is stranded `running` and all media processing halts until restart | reliability | Unguarded `constantize` can kill the scheduler run-loop permanently |

### Found during implementation (not audit findings)

| Item | Description | Type | Summary |
|------|-------------|------|---------|
| FI-01 | `edited_at`/`disabled_at` come back as `+0700` instead of UTC across note_feed/note_comment/project_member specs — same wall-clock digits, a 7-hour instant shift. Systemic (3 services, 2 columns) → likely global Sequel/`TZ` config. Confirm test-only vs. real write discrepancy before fixing | test / config | `Time`-typed columns round-trip in the wrong timezone |
| FI-02 | Rows created by `media_repo.create` sometimes survive into the next example. **Update:** a leaked `test_resource/test_key` row is now persistently in `idah_media_test`, so the media suite fails *deterministically* until the DB is reset. Root cause (Verse/Sequel `rollback_on_exit` gap) still unidentified | test isolation | `medias` specs leak rows across examples despite the per-example transaction |

### DIFF
--- a/app/dataset/app/expo/annotations_expo_spec.rb
+++ b/app/dataset/app/expo/annotations_expo_spec.rb
@@ -112,10 +112,6 @@ RSpec.describe AnnotationsExpo, type: :exposition, as: :system do
       }
     end
 
-    def rpc_batch_call(calls)
-      post "/annotations/rpc", calls, { "CONTENT_TYPE" => "application/json" }
-    end
-
     it "show" do
       expect(service).to receive(:show).with(uuid).and_return(annotation_record)
       rpc_call rpc_cmd("show", { id: uuid })
--- a/app/dataset/app/model/entry.rb
+++ b/app/dataset/app/model/entry.rb
@@ -238,15 +238,27 @@ module Entry
           entry = find!(id)
 
           if entry.status == "pending"
-            update!(
+            account_id = auth_context.metadata[:id]
+
+            # Narrowed scope makes the claim atomic: a concurrent claimer's
+            # UPDATE re-evaluates the condition after the winner commits,
+            # matches zero rows and loses deterministically instead of
+            # silently overwriting the first claim.
+            updated = update(
               id,
               {
-                assigned_to_id: auth_context.metadata[:id],
+                assigned_to_id: account_id,
                 assigned_to_email: auth_context.metadata[:email],
                 status: "in_progress"
               },
               scope: scoped(:read)
+                     .where(status: "pending")
+                     .where(Sequel.expr(assigned_to_id: nil) | Sequel.expr(assigned_to_id: account_id))
             )
+
+            unless updated
+              raise Verse::Error::Unauthorized, "Entry already claimed by another user"
+            end
           end
         end
       end
--- a/app/dataset/app/model/entry.rb
+++ b/app/dataset/app/model/entry.rb
@@ -323,7 +323,8 @@ module Entry
     end
 
     def mark_entries_status_as(job_id, status)
-      entry = find_by!({ job_id: job_id, status: "processing" })
+      entry = find_by({ job_id: job_id, status: "processing" })
+      return unless entry
 
       transaction do
         update!(entry.id, { status: })
--- a/app/dataset/app/model/entry_spec.rb
+++ b/app/dataset/app/model/entry_spec.rb
@@ -9,5 +9,85 @@ RSpec.describe Entry, database: true do
     it "can be instantiated" do
       expect(subject).to be_a(Entry::Repository)
     end
+
+    describe "#select" do
+      let(:user_a_context) do
+        Verse::Auth::Context.from_role("admin", metadata: { id: 100, email: "user_a@example.com" })
+      end
+
+      let(:user_b_context) do
+        Verse::Auth::Context.from_role("admin", metadata: { id: 200, email: "user_b@example.com" })
+      end
+
+      let(:repo_a) { described_class.new(user_a_context) }
+      let(:repo_b) { described_class.new(user_b_context) }
+
+      let!(:project_id) do
+        Project::Repository.new(user_a_context).create(
+          name: "Test Project",
+          description: "A test project",
+          created_by_email: "user_a@example.com",
+          organization_id: 1,
+        )
+      end
+
+      let!(:dataset_id) do
+        Dataset::Repository.new(user_a_context).create(
+          modality: "image_labeling",
+          labels: ["cat", "dog"],
+          labeling_configuration: { "width" => 100, "height" => 100 },
+          workflow_configuration: {},
+          project_id:
+        )
+      end
+
+      let(:entry_attributes) do
+        {
+          priority: 1,
+          wf_step: "start",
+          status: "pending",
+          project_id:,
+          dataset_id:
+        }
+      end
+
+      it "claims an unassigned pending entry" do
+        entry_id = repo_a.create(entry_attributes)
+
+        repo_b.select(entry_id)
+
+        entry = repo_b.find!(entry_id)
+        expect(entry.assigned_to_id).to eq(200)
+        expect(entry.assigned_to_email).to eq("user_b@example.com")
+        expect(entry.status).to eq("in_progress")
+      end
+
+      it "re-selects a pending entry already assigned to the caller" do
+        entry_id = repo_a.create(
+          entry_attributes.merge(assigned_to_id: 100, assigned_to_email: "user_a@example.com")
+        )
+
+        repo_a.select(entry_id)
+
+        entry = repo_a.find!(entry_id)
+        expect(entry.assigned_to_id).to eq(100)
+        expect(entry.status).to eq("in_progress")
+      end
+
+      it "does not steal a pending entry already claimed by another user" do
+        entry_id = repo_a.create(
+          entry_attributes.merge(assigned_to_id: 100, assigned_to_email: "user_a@example.com")
+        )
+
+        expect {
+          repo_b.select(entry_id)
+        }.to raise_error(Verse::Error::Unauthorized, /already claimed/)
+
+        entry = repo_a.find!(entry_id)
+        expect(entry.assigned_to_id).to eq(100)
+        expect(entry.assigned_to_email).to eq("user_a@example.com")
+        expect(entry.status).to eq("pending")
+      end
+    end
   end
 end
--- a/app/dataset/app/model/entry_spec.rb
+++ b/app/dataset/app/model/entry_spec.rb
@@ -89,5 +89,70 @@ RSpec.describe Entry, database: true do
         expect(entry.status).to eq("pending")
       end
     end
+
+    describe "#custom_filter :assigned" do
+      let(:system_auth_context) { Verse::Auth::Context[:system] }
+
+      subject { described_class.new(system_auth_context) }
+
+      let!(:project_id) do
+        Project::Repository.new(system_auth_context).create(
+          name: "Filter Project",
+          description: "A test project",
+          created_by_email: "admin@example.com",
+          organization_id: 1,
+        )
+      end
+
+      let!(:dataset_id) do
+        Dataset::Repository.new(system_auth_context).create(
+          modality: "image_labeling",
+          labels: ["cat", "dog"],
+          labeling_configuration: { "width" => 100, "height" => 100 },
+          workflow_configuration: {},
+          project_id:
+        )
+      end
+
+      let!(:assigned_entry_id) do
+        subject.create(
+          priority: 1,
+          wf_step: "start",
+          status: "in_progress",
+          assigned_to_id: 4,
+          assigned_to_email: "annotator@example.com",
+          project_id:,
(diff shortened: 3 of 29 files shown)


rcserrano06:
Task linked: [CU-869e0fx4d Dataset and Media (jobs/processor/misc)](https://app.clickup.com/t/869e0fx4d)

PR #535, opened 2026-07-09
Frontend: add unique tiebreaker to sorting to prevent missing annotations

rcserrano06:


rcserrano06:
Task linked: [CU-869e0h97p IDAH - Annotation count not synced across browsers/computers](https://app.clickup.com/t/869e0h97p)

PR #544, opened 2026-07-16
IDAH Disable Save Changes Button On Edit Modal When No Changes Made

FirstKarnravee:
## Summary
Disables the "Save Changes" button on Edit modals until the user actually changes something, and re-disables it if all fields are reverted to their original values. Also hardens the dirty-state implementation (single source of truth for editable fields, edit-state isolation).

## Why
On any Edit modal the "Save Changes" button was always clickable even with no changes, offering no feedback about whether there was anything to save and allowing pointless save calls. (ClickUp 869dv5nfy)

## Changes
**Core feature**
- Each edit modal tracks a `savedSnapshot` (original record) vs `editedSnapshot` (current form values) and drives the shared `FormModal.disabled` on `hasUnsavedChanges` (update action only; create modals unchanged).
- `FormModal` Enter-to-submit now respects `disabled`, so a clean form can't be saved via the Enter key.
- Covered modals: project, dataset, organization, api-key, account.

**Correctness fixes**
- **Account / `sso_channel`:** the field is displayed but read-only and not emitted by the form. It was in the dirty diff, which made the button appear enabled on open for SSO accounts and could null the value on save. Excluded from the diff and no longer overwritten on save.
- **Dataset / "Copy label configurations from":** selecting a source now marks the form dirty (it changes the saved label config). The select can't be cleared, so there's no revert-to-none for it.

**Hardening**
- **Single source of truth for editable fields:** `serializeEditableFields` is reused for both the original-record and current-value snapshots, so the two sides can't drift (root cause of the `sso_channel` bug). Keys are limited to what the form emits via `onValueChange`.
- **Edit-state isolation (`Record.clone()`):** modals edit a clone instead of mutating the passed-in record in place, so unsaved keystrokes no longer leak into parent-held references (e.g. the project delete toast).

## Testing
- `npm run check` / prettier / eslint on touched files — no new errors (pre-existing read-only-assignment / import errors remain).
- Manual, per edit modal: open → disabled; change a field → enabled; revert → disabled; press Enter with no change → does not submit; reopen after cancel → disabled.
- Edge cases: SSO account opens disabled; dataset copy-from selection enables Save; saving each entity succeeds unchanged; edit-then-cancel a project name then delete → toast shows the original name.

## Notes
- Create modals are intentionally unchanged (out of scope).
- `Record.clone()` copies type + id + attributes only; it deliberately drops relationship/include caches, so it's for attribute-level editing.


rcserrano06:
Task linked: [CU-869dv5nfy IDAH - Disable "Save Changes" button on Edit modal when no changes made](https://app.clickup.com/t/869dv5nfy)

jiramarc (approved):

rcserrano06 (approved):

PR #545, opened 2026-07-17
Page freezes after deleting a configuration in label editor

jiramarc:
…figuration

Deleting a non-selected label configuration incorrectly cleared the selected config key, causing a null dereference crash that froze the page. The fix only reassigns the selection when the deleted config was the one selected.

Removed the now-dead labelConfigIsEmpty derived check and hardened render guards to check the actual selectedLabelConfig instead of a whole-config-empty condition.

Refs: CU-869e5vc8p

### DIFF
--- a/app/frontend/src/lib/components/app/datasets/labels/label-config-editor.svelte
+++ b/app/frontend/src/lib/components/app/datasets/labels/label-config-editor.svelte
@@ -106,7 +106,6 @@
       shapeDrag.end();
     },
   };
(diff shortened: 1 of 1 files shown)


rcserrano06:
Task linked: [CU-869e5vc8p IDAH - Page freezes after deleting a configuration in label editor](https://app.clickup.com/t/869e5vc8p)

rcserrano06 (approved):

PR #551, opened 2026-07-23
Deleting the only keyframe of an annotation doesn't remove the annotation

jiramarc:
## What
Deleting the only keyframe of a video annotation now removes the whole annotation instead of leaving an invalid zero-keyframe annotation in the timeline.

## Why
An annotation with zero keyframes is invalid and should not exist. Previously it lingered in the timeline, hurting data quality and downstream review/export.

## How
- `keyframe_delete` command: when the last keyframe is removed, it now deletes the whole annotation (undo recreates it), reusing the same primitives as `annotation.delete`, instead of writing back an empty shape.
- Timeline context menu (`_TrackBlockContextMenu.svelte`): hides the "Delete keyframe" action for single-keyframe annotations, steering users to "Delete annotation".

## Testing
- `npm run check` → 0 errors
- `npm test` → 277 passed
- Manual: single-keyframe delete via context menu and via Ctrl+Delete removes the whole annotation; undo restores it; deleting a non-last keyframe on a multi-keyframe annotation is unaffected.

---
Closes CU-869e8azzy

### DIFF
--- a/plugins/idah-video/frontend/src/lib/commands/annotation/keyframe_delete.ts
+++ b/plugins/idah-video/frontend/src/lib/commands/annotation/keyframe_delete.ts
@@ -77,13 +77,26 @@ export function register(driver: IIdahDriverV2): void {
       if (idx === -1) return noopAction(command);
 
       const snapshot: AnnotationItem = { ...record, shape: { ...record.shape, frames: [...frames] } };
+      // Removing the last remaining keyframe must delete the whole annotation —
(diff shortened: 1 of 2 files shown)


rcserrano06:
Task linked: [CU-869e8azzy IDAH - Deleting the only keyframe of an annotation doesn't remove the annotation](https://app.clickup.com/t/869e8azzy)

rcserrano06 (approved):

PR #554, opened 2026-07-30
IDAH Update Delete Confirmation Dialogs

FirstKarnravee:
## Summary

Replaces 18 locally-rendered `<ConfirmModal>` instances across 15 files with a single renderer mounted once in `src/routes/+layout.svelte`. Callers now ask for a confirmation through `showConfirmModal()` and hold no modal state of their own.

The original scope of this branch — naming the record in delete dialogs — is included; the refactor grew out of it.

## Why

Every call site owned an `open` boolean, its own `<ConfirmModal>` markup, and its own close-timing policy. That drifted:

- **Escape never ran cancel cleanup.** `ConfirmModal` passed only `bind:open`, no `onOpenChange`, so `onCancel` never fired on Escape. Three sites used it purely for cleanup and were left with stale state.
- **Two dialogs never closed** after a successful confirm (`organization-owner-row-action-cell`, `project-member-role-cell`).
- Five of the call sites are data-table row cells, so a 50-row table held ~100 modal component instances.

## Approach

A single `<ConfirmModal />` in the root layout, driven by a module-scoped service.

Lifecycle is **hybrid, selected per action by the presence of `run`** — there is no mode flag:

```ts
// sync — closes on click
const choice = await showConfirmModal({ title, description, confirmLabel: "Yes, Overwrite" });
if (choice === ConfirmModalChoice.Cancel) return;

// async — stays open, shows pending, closes only on success
await showConfirmModal({
  title, description,
  onConfirm: async () => {
    try { await deleteThing(); showToast.success(…); }
    catch (error) { showToast.error(…); return ConfirmModalResult.KeepOpen; }
  },
});
```

Failure is signalled by a **return value, not an exception**, matching the convention already used by `entry-actions.ts` ("false on failure, toast already shown"). A thrown error is caught, logged, and treated as `KeepOpen` — a safety net so an unhandled rejection cannot silently close a destructive modal.

16 of the 18 call sites use async mode, preserving today's stay-open-on-error behaviour. `labels/+page.svelte` mixes both in one dialog: *Don't Save* is sync, *Save* keeps the prompt open when the save fails.

`run` must not navigate — it executes while the modal is open, where a `goto()` can trip a navigation guard. The four sites that navigate do so after the `await`.

## Behaviour changes

**Fixes**
- Escape now runs cancel cleanup everywhere.
- The two never-closing dialogs now close on success and stay open on failure.
- Bulk dataset delete read `"delete dataset 3 datasets?"` — fixed, now uses the existing `pluralizeUnit` helper.
- The bulk-delete success toast reported **"The 0 datasets have been deleted"**: `selectedDatasetsCount` is derived from a store that is cleared before the toast fires. The count is now captured up front.
- Dataset tabs no longer move optimistically. `<Tabs bind:value={activeTab}>` wrote back into a `$derived`, so cancelling the unsaved-changes prompt left the tab on a page the router never reached. The tab now follows the URL, and keyboard activation navigates instead of only moving the highlight.

**Intentional UX copy change — please review**
- "This action cannot be undone." added to the bulk dataset-delete dialog. This is new copy, not a migration artifact.

**Behaviour deliberately preserved**
- Stay-open-on-error on all 16 destructive actions.
- `note-dropdown-menus` closes on failure, as it did before via `try/finally` — the one place a `catch` intentionally omits `KeepOpen`, commented in place.

## Concurrency

Confirmations are strictly sequential; an overlapping `showConfirmModal()` throws. Two details worth a reviewer's attention:

- The guard checks "is someone awaiting a decision?" (`activeResolve`), **not** "is a modal on screen?" (`request`). `request` outlives the promise so the closing modal has something to render, so gating on it rejected a legitimate `await show(A); show(B)` for the length of the exit animation.
- `{#key request.id}` on the content payload is load-bearing *because* of that: a request can now replace another without passing through `null`, and without the key Svelte would reuse the previous content instance along with its in-flight fetch.

`labels/+page.svelte`'s `beforeNavigate` guard is re-entrant — the overlay blocks clicks but not the browser Back button — so it bails via `isConfirmModalActive()` rather than calling `showConfirmModal()` twice.

## Testing

31 unit/integration tests (23 service, 8 component), covering both lifecycle modes, `KeepOpen`, thrown-`run` handling, pending state, stale-result suppression, sequential confirms mid-exit-animation, and the SSR guard.

Manually verified in Chrome against the real components: dialog stacking and focus return over an open `Sheet`, Escape behaviour, and all six unsaved-changes navigation flows including browser Back.

`svelte-check` is unchanged at the repo baseline (186 errors / 44 files — all pre-existing). ESLint and Prettier are clean on all 26 files this branch touches; the 27 ESLint errors reported repo-wide are in `ui/` and `data/` files this branch does not touch. The 13 failing tests in the full suite are pre-existing and were verified against a clean worktree at `main`.

## Out of scope

Deliberately excluded, with reasoning: `FormModal` (29 files, holds form state — does not belong in a module singleton), the same optimistic-tab issue in the organizations/projects layouts (keyboard-only, no navigation guard there, so no user-visible bug), `AbortSignal` support for `run`, and `project-dataset-dropdown-menu`'s `datasetRecord?.name` which can render `"undefined"`.

## Notes for reviewers

- A `no-restricted-imports` ESLint rule enforces that only `src/routes/+layout.svelte` may import the component.
- `showConfirmModal` is a no-op under SSR — it logs in dev and resolves as cancelled.
- The bulk-delete copy addition above is the one change that is not behaviour-preserving by design.

ClickUp: CU-869ebfnc5

### DIFF
--- a/app/frontend/src/lib/components/app/overlays/modals/confirm-modal.types.ts
+++ b/app/frontend/src/lib/components/app/overlays/modals/confirm-modal.types.ts
@@ -1,12 +1,100 @@
-import type { Snippet } from "svelte";
+import type { Component } from "svelte";
 
-import type { ModalBaseProps } from "@/components/app/overlays/modals/Modal.types";
+import type { ButtonVariant } from "@/components/ui/button/button.svelte";
 
-export interface ConfirmModalBaseProps extends ModalBaseProps {
+/** Universal outcome shared by every confirm modal. */
+export const ConfirmModalChoice = {
+  Confirm: "confirm",
+  Cancel: "cancel",
+} as const;
+export type ConfirmModalChoice = (typeof ConfirmModalChoice)[keyof typeof ConfirmModalChoice];
+
+/**
+ * Returned by an action's `run` to control the modal. Returning nothing means success and
+ * the modal closes.
+ *
+ * Mirrors the convention already used by `entry-actions.ts` ("false on failure, toast
+ * already shown"): failures are reported by the caller and signalled by a return value,
+ * never by throwing.
+ */
+export const ConfirmModalResult = {
+  KeepOpen: "keep-open",
+} as const;
+export type ConfirmModalResult = (typeof ConfirmModalResult)[keyof typeof ConfirmModalResult];
+
+/**
+ * Async work owned by an action.
+ *
+ * Runs while the modal stays open in a pending state. Return `ConfirmModalResult.KeepOpen`
+ * to keep it open (the caller has already shown its own error toast); return nothing to
+ * close. Throwing is treated as `KeepOpen` and logged — it is a safety net so an unhandled
+ * rejection cannot silently close a destructive modal, not the failure channel.
+ *
+ * `run` must not navigate: it executes while the modal is open, so a `goto()` here can
+ * trigger a navigation guard that opens a second modal. Navigate after `showConfirmModal()`
+ * resolves instead.
+ */
+export type ConfirmModalRun = () => Promise<ConfirmModalResult | void> | ConfirmModalResult | void;
+
+export interface ConfirmModalAction<Id extends string = string> {
+  /** Returned by `showConfirmModal()` when this action is chosen. */
+  id: Id;
+  label: string;
+  variant?: ButtonVariant;
+  /** Present selects async mode for this action; absent closes the modal immediately. */
+  run?: ConfirmModalRun;
+  /** Button label while `run` is in flight. Defaults to the action's own label. */
+  pendingLabel?: string;
+}
+
+export interface ConfirmModalContent {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- payload props are caller-defined
+  component: Component<any>;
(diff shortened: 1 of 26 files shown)


rcserrano06:
Task linked: [CU-869ebfnc5 IDAH - Update delete confirmation dialogs](https://app.clickup.com/t/869ebfnc5)

app/frontend/src/lib/components/app/overlays/modals/confirm-modal.types.ts
```diff
@@ -1,12 +1,100 @@
...
+
+/**
+ * Returned by an action's `run` to control the modal. Returning nothing means success and
+ * the modal closes.
+ *
+ * Mirrors the convention already used by `entry-actions.ts` ("false on failure, toast
+ * already shown"): failures are reported by the caller and signalled by a return value,
+ * never by throwing.
+ */
+export const ConfirmModalResult = {
```
jiramarc: Why constants is starting with capital letter?

jiramarc (approved):

rcserrano06 (approved):

PR #558, opened 2026-07-30
Update updcli exporter return

Yraimtj:
- with this we can get entry name from updcli entry show, and entry id for updcli annotation show
- we can also send on the updcli-static annotation create --shape argument as a file to avoid this Error `Errno::E2BIG: Argument list too long`. it is cause by the fact when dealing with mask shape, the dimension value send to the --shape argument is a long long string. so the command is failed. So i add the capability to send the string in a file with '@' on the file_path


rcserrano06:
Task linked: [CU-869ebk4dx IDAH - Bitmask Annotation Import/Export Scripts](https://app.clickup.com/t/869ebk4dx)

anykeyh (changes requested):
I don't get this code at all

anykeyh:
This is not good; there is enough information to map annotation to media already; adding a header in the media URL feels hacky; the header naming doesn't follow convention from other headers etc... You can write your script without this hack.

app/sync/app/service/exports/upd/exporter.rb
```diff
@@ -111,6 +111,18 @@ def append_entry(file_path, dataset_id, entry, include_medias)
           )
         )
 
+        # Fetch original media metadata to get original width and height
+        original_media = entry.medias({ key: "" }).first
+        if original_media
+          media_meta = original_media.record.data[:attributes][:meta] || {}
+          if media_meta[:width]
```
anykeyh: You could add all meta data from media:

```ruby
# Inject; normalize = `-` + capital case
media_meta.each{ |k, v| metadata[normalize(k)] = v }
```

anykeyh (changes requested):

anykeyh (approved):

PR #555, opened 2026-07-30
Editor Settings menu

sorakrit-audi:

PR #566, opened 2026-08-10
IDAH - Display Annotation Category Label

jiramarc:
Shows an annotation's category name on the canvas, with a three-mode visibility setting in the toolbar Settings popover.

ClickUp: [869eehq8w](https://app.clickup.com/t/869eehq8w) · [Figma](https://www.figma.com/design/58gVy4PHhvhPUsK8iELvCR/IDAH---Design-System?node-id=2134-1845)

> [!NOTE]
> **Base is #555, not `main`.** This branch was cut from `features/CU-869e6qan6/opacity-settings` and extends its Settings menu. Based on `main` the diff would include #555's unmerged work. #555's tip has been merged in, so this is up to date with it and currently conflict-free.

## What it does

- Category name renders at the **top-left** of each annotation, white text with a dark outline, over the media.
- Visibility is a tri-state toolbar setting: **On** / **On hover** / **Off**, default **Off**.
- Works for every shape except **Bitmask**, which is drawn to a canvas layer rather than into the shapes SVG.
- Both plugins: `idah-image` (bbox, polygon, line, circle, ellipse) and `idah-video` (bbox, polygon).

## Commits

| Commit | Subtask |
|---|---|
| `ad14f4b4` category label visibility setting | [869ef2pf9](https://app.clickup.com/t/869ef2pf9) |
| `9b648dd9` hovered annotation store; wire shape hover handlers | [869ef2pfp](https://app.clickup.com/t/869ef2pfp) |
| `f1d02a3a` label anchor geometry and svg text wrapping | [869ef2pg2](https://app.clickup.com/t/869ef2pg2) |
| `dedec78a` render category labels on canvas | [869ef2ph3](https://app.clickup.com/t/869ef2ph3) |
| `b29f7f6f` idah-video; mirror category label feature | [869ef2pht](https://app.clickup.com/t/869ef2pht) |
| `f9689695` ignore local CLAUDE.md | — |
| `23f9595a` merge opacity-settings into category-label | — |

> [!NOTE]
> Account-level persistence was prototyped on this branch and then **reverted** (commit `073313a2a`) — it is being reimplemented on a dedicated sub-branch using an `upsert` API. This PR is the display feature only; the setting is session-only here (see below).

## Merge with #555 — worth a look

Both branches edited each plugin's `settings.ts`. The resolution keeps **every** setting from both sides — their `render-mode` / `color-mode` (and `time-display` on video) plus `label-visibility` from here. Their side had also independently deleted the `mock-options` placeholder.

Their descriptor contract changed, so `label-visibility` was adapted: `default` was dropped from `ISettingItemBase`, `description` was added, and the `set: (v: number)` typing landed on both sides independently.

**One judgment call:** their options settings route writes through `driver.command.call(...)` so a shortcut and the popover share one mutation path. `label-visibility` sets `ui` directly instead — it has no shortcut or palette entry, so the popover is the only writer and core's renderer calls `settings.emitChange()` itself after `set()`. If a shortcut is ever added, convert it to a command.

## Notable decisions

Full rationale for all 18 decisions is in the [decision-log comment](https://app.clickup.com/t/869eehq8w). The ones most likely to be questioned:

- **Labels render as one layer after all shapes**, not per-shape. Shapes are z-sorted, so a per-shape label gets painted over by later shapes' fills — worst exactly in the dense overlapping case where labels matter most.
- **SVG `<text>` + `<tspan>`, not `<foreignObject>` or an HTML overlay.** Both would give free CSS wrapping, but `paint-order: stroke` — which produces the outline in the design — is SVG-only.
- **Anchor is the AABB of the *rotated* corners.** One rule for every shape at every angle. Rejected: welding to a fixed corner (drifts off "top-left" at large angles) and always-topmost-left (snaps corner-to-corner mid-rotation).
- **No coloured chip.** A light category colour on a light photo is unreadable; shape colour already carries category identity.
- **Session-only persistence**, matching opacity. localStorage was offered and declined; account-level persistence is being delivered separately on its own sub-branch via an `upsert` account-settings API (tracked under 869ef2pj0).

## Two traps worth knowing

Both documented in code comments and covered by regression tests:

- **`categoryValueToLabel(value)` with one argument POPS the leaf segment** — it's a parent-path helper. Omitting the second argument silently truncates every label to `Vehicle / Commercial`.
- **An ellipse is stored as `[[cx, cy], [rx, ry]] + angle`, not as 4 AABB corners.** The corner form exists only inside `EllipseShape` for editing. I got this wrong initially and no ellipse rendered a label; the unit tests missed it because they were written against the same wrong model. Fixed, with an explicit `does NOT require the 4-corner AABB form` test.

## Testing

| | Tests | `svelte-check` | Build | Manual QA |
|---|---|---|---|---|
| `idah-image` | 464 pass (+46) | 0 errors | ✅ | ✅ full checklist |
| `idah-video` | 335 pass (+29) | 0 errors | ✅ | ✅ full checklist |

Manual QA covered both plugins end to end: all three modes, zoom range, rotation, wrapping and overflow, click-through, hidden annotations, non-regression on every modified shape component, and for video the keyframe-specific cases (interpolated scrubbing, frame-range entry/exit, playback tracking). Checklists are on the ticket.

`svelte-check` on `idah-image` went from **3 pre-existing errors to 0** — the setting descriptors' `set` handlers were implicit-`any`.

> [!WARNING]
> **Unrelated, but worth flagging for #555:** `app/frontend` has 13 failing tests on that branch — `BackendDataSource`, `jsonrpc`, `KeyMapBuilder`, and a broken import in `page.svelte.test.ts` (`./(app)/+page.svelte` does not resolve). Not caused by this branch: `app/` and `common/` here are a 0-line diff against `origin/features/CU-869e6qan6/opacity-settings`. This workflow does not run them, so they are not gating either PR.

## Known limitations — by design

- **Overlapping labels are not de-duplicated.** Default is Off and "On hover" shows one at a time, so this is opt-in; collision layout wasn't worth building before seeing whether it hurts.
- **No clamping at media edges** — labels overflow onto the workspace background rather than detaching from their anchor.
- **No Bitmask labels.**
- **Setting resets on reload** (session-only; account-level persistence lands on the `upsert` sub-branch).

## Follow-ups filed

- [869ef2pj0](https://app.clickup.com/t/869ef2pj0) — account-level persistence (now in progress on a dedicated `upsert` sub-branch). Also: the ticket's acceptance criteria contradict its body on the default mode (body says Off, AC says Auto; **Off** is implemented) — needs correcting or QA will file it as a bug — plus collision-handling work and four stale code comments (two in plugin `ui.test.ts` headers, two in core describing an obsolete type-mirroring recipe).


rcserrano06:
Task linked: [CU-869e6qan6 IDAH - Opacity settings for annotation filled color and media](https://app.clickup.com/t/869e6qan6)
Task linked: [CU-869eehq8w IDAH - Display Annotation Category Label](https://app.clickup.com/t/869eehq8w)

app/frontend/src/lib/plugin/IdahPlugin.svelte
```diff
@@ -60,14 +60,28 @@
...
+     *
+     * (2) the plugin reads its persisted settings (e.g. category label visibility)
+     * synchronously in init(), and reactivity does not cross the plugin/core
+     * bundle boundary — so the value must already be present.
+     *
+     * Guarded so a settings failure never blocks plugin startup.
+     */
+    driver.accountSettings
+      .load($authStatus.authContext?.id ?? "")
+      .catch(() => {})
```
rcserrano06: Should show an error message when there is an issue loading the account settings

app/setting/app/service/account_settings/defaults.rb
```diff
@@ -2,13 +2,18 @@
...
+      { key: "notification:organization:activities", plugin: "", value: true },
+      { key: "notification:project:activities", plugin: "", value: true },
       # command palette shortcut overrides — { "<command-name>" => "<shortcut>" }
-      "command:shortcut" => {},
-    }.freeze
+      { key: "command:shortcut", plugin: "", value: {} },
+      # annotation category label visibility — "always" | "hover" | "never".
+      # Stored per plugin so image and video preferences stay independent.
+      { key: "annotation:category.label-visibility", plugin: "idah-image", value: "never" },
+      { key: "annotation:category.label-visibility", plugin: "idah-video", value: "never" },
```
rcserrano06: I think this should be plugin specific settings and should not be directly added on the settings service.

It might be best also to re-think how account settings should be set on the backend, I would recommend adding default value when registering the setting on the Plugin which will allow the setting to not be saved yet on the backend, then only when it's changed it will be set on the backend.

rcserrano06 (changes requested):

rcserrano06 (approved):

PR #567, opened 2026-08-10
IDAH - Multi-select annotation

Yraimtj:
what we can do.

- select multiple shape in the viewport by clicking in shape with shift + click, add shape if not already selected, remove from the selection if already selected
- can also make selection by drawing rectangle with shift + drag
- as we already had this shift + drag action before for polygon vertex selection , i change to be alt + drag (it only for polygon selected)
- can also make multi selection on the group of annotation on the time line with shift + click
- when shape is selected in viewport, correspondg annotation on the timeline is also highlighted. 
- can delete multiple selected shape annotation with short cut keyboard delete
- can move already multiple shape
- can make copy and paste action too. we paste on the last  position of the cursor
- and we can extend annotation depending on which annotation you click on the track
- if you click on right of all the selected annotation => extend all previous to frame n°
- if you click on left of all the selected annotation => extend all next to ...
- if the one you clicked have blank in middle the two extend will appear so it up to the user to decide where he want to extend
- if one of the selected group annotation is locked : warning will appear to say can not delete
- for extend all, if one group annotation is locked it will not be extend, other will be extend


rcserrano06:
Task linked: [CU-869dgvcun IDAH - Multi-select annotation](https://app.clickup.com/t/869dgvcun)

plugins/idah-video/frontend/src/lib/commands/annotation/extend_all_next.ts
```diff
@@ -0,0 +1,178 @@
...
+            driver.command.call("annotation.keyframe_add", {
+              annotationId: nextAnn.id,
+              selection: {
+                frame: cappedFrame,
+                ...nearest,
+              },
+            });
+          }
+        },
+        // No undo — the nested keyframe_add handles its own undo.
```
brian-west-ingedata: I dont think we should call command from commands ( driver.command.call( "annotation.keyframe_add" ))
if we extend all on 10 annotations undo should undo those 10 in one go, not only the last one made by those inner call.

plugins/idah-video/frontend/src/lib/commands/annotation/extend_all_prev.ts
```diff
@@ -0,0 +1,174 @@
...
+            driver.command.call("annotation.keyframe_add", {
+              annotationId: prevAnn.id,
+              selection: {
+                frame: cappedFrame,
+                ...nearest,
+              },
+            });
+          }
+        },
+        // No undo — the nested keyframe_add handles its own undo.
```
brian-west-ingedata: [same](https://github.com/idah-ai/idah/pull/567/changes#r3818240123)

plugins/idah-video/frontend/src/lib/components/App/Viewport/VideoController.svelte
```diff
@@ -206,10 +206,9 @@
...
         disabled={disabledSplitButton}
         onclick={() => {
-          const ann = selection.value?.type === "annotation" ? (selection.value as any).annotation : undefined;
-          if (ann)
+            // No annotationId — the split command will batch-split
+            // all selected annotations when called without opts.
             getDriver().command.call("annotation.split", {
-              annotationId: ann.metadata?.id ?? ann.id,
               at: viewport.video.currentFrame.value,
             });
```
brian-west-ingedata: it feels off to me to depend on internal state from within command, they should inject their dependencies. unit testable, prevent any race condition, etc..

brian-west-ingedata (commented):
we could probably clean some of those console.log

rcserrano06 (changes requested):
## Issues
1. After moving multiple selected annotations, the undo triggers per annotation which should be for all annotations, might be good to have multiple update command.
2. Error when copy -> paste -> undo -> redo -> undo.
3. When I copy annotation on frame 1 and paste it on frame 2 the annotation produce no frames and start(frame 1) > end(frmae 0).

4. Shift drag select multiple cannot start on top of annotation.
5. Shift drag select multiple exits when mouse is outside viewport and selects the texts on the right sidebar.
6. When I move multiple selection and mouse went outside the viewport the selection exits and move only specific annotations.

## Improvements
1. Improve the selected annotation during multi select by adding gray points on the vertices which is not movable, for bbox remove rotation handle.

rcserrano06 (approved):

PR #577, opened 2026-08-24
IDAH - Refactor Account Settings

jiramarc:


rcserrano06:
Task linked: [CU-869em9z17 IDAH - Refactor Account Settings](https://app.clickup.com/t/869em9z17)

rcserrano06:
Task linked: [CU-869eehq8w IDAH - Display Annotation Category Label](https://app.clickup.com/t/869eehq8w)

app/frontend/src/lib/plugin/v2/driver/manager/account-settings-manager.svelte.ts
```diff
@@ -25,8 +26,20 @@ export class AccountSettingsManager {
...
+    return `${plugin} ${key}`;
+  }
+
+  get(key: string, plugin = ""): SettingValue | undefined {
+    return this.settings.get(this.#mapKey(key, plugin))?.value;
+  }
+
+  // Create-or-update a setting by (key, plugin) and cache the returned record.
+  // The single generic write path exposed to plugins and core UI.
+  async upsert(key: string, value: SettingValue, plugin = ""): Promise<void> {
```
rcserrano06: I think `plugin` should be retrieve from the driver and not as arguments here as this will allow plugin developers to just declare any plugin and be able to update the settings of other plugins.

Same with the `get` method

rcserrano06 (changes requested):

rcserrano06:
Task linked: [CU-869e6qan6 IDAH - Opacity settings for annotation filled color and media](https://app.clickup.com/t/869e6qan6)

rcserrano06 (approved):

app/frontend/src/lib/plugin/v2/types.ts
```diff
@@ -674,8 +676,17 @@ export interface IAccountSettingsDriverV2 {
...
+   * Read a setting in the active plugin's namespace, or undefined if not
+   * loaded.
+   */
+  get(key: string): AccountSettingValue | undefined;
+
+  /**
+   * Create-or-update a setting in the active plugin's namespace. The row is
+   * created on first write and updated afterward.
+   */
+  upsert(key: string, value: AccountSettingValue): Promise<void>;
```
rcserrano06, line 691: I think we can refactor `get` and `set` to this:
```ts
set(key: string, value: unknown): void {
  this.settings.set(key, value);
}

get<T>(key: string): T | undefined {
  return this.settings.get(key) as T | undefined;
}
```
`set` stores anything without needing a type argument. `get` needs <T> specified explicitly because TypeScript has no value to infer it from. Without it, T defaults to unknown.

Usage
```ts
settings.set('fontSize', 14);
const fontSize = settings.get<number>('fontSize'); // typed as number | undefined
```

rcserrano06 (changes requested):

rcserrano06 (approved):

PR #594, opened 2026-09-08
Plugins: idah-image refactor readOnly condition in shape handlers

rcserrano06:


rcserrano06:
Task linked: [CU-869ek6g4t IDAH - Multi-select and duplicate annotation features for image plugin](https://app.clickup.com/t/869ek6g4t)

PR #596, opened 2026-09-08
Plugins: idah-video fix selection.copy to copy only selected annotation

rcserrano06:


rcserrano06:
Task linked: [CU-869dgvcun IDAH - Multi-select annotation](https://app.clickup.com/t/869dgvcun)

PR #601, opened 2026-09-14
IDAH - Add capability to extend workflow from externat app plugins implementation

Yraimtj:
- add callback endpoint so external app call send request to idah to move entry to next step
- when create dataset if need external app step configure the dataset creation to add app url link and callback token
- add qc workflow plugins example


rcserrano06:
Task linked: [CU-869cr63pk IDAH - Workflow registry and workflow selection](https://app.clickup.com/t/869cr63pk)

